PR #17, opened 2022-10-25
test: ordered list starting high enough to overflow u8

ChristianMurphy:
```
panicked at 'expected list value up to u8: ParseIntError { kind: PosOverflow }', //micromark-rs/src/to_mdast.rs:1427:6
```

Cause by the expectation here:
https://github.com/wooorm/markdown-rs/blob/cd782d3c3563ef711905429c6d9e1a88acf14fd7/src/to_mdast.rs#L1420-L1427

### DIFF
--- a/tests/fuzz.rs
+++ b/tests/fuzz.rs
@@ -1,5 +1,10 @@
 extern crate markdown;
-use markdown::{to_html, to_html_with_options, Options};
+use markdown::{
+    mdast::{List, ListItem, Node, Root},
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
🤔 is this a develop mode only assertion? I don't seem to be able to trigger the same exception in the test suite

wooorm:
nah, difference between mdast and html impls

ChristianMurphy:
Doh! Completely missed that, thanks!
Updating the test now.

PR #21, opened 2022-10-26
test: links with tilde or star as label

ChristianMurphy:
triggers
https://github.com/wooorm/markdown-rs/blob/b910e37bb387a62509a7fda26617e9870ed6d56f/src/to_mdast.rs#L155-L167


ChristianMurphy:
replicated issue in commonmark as well, updated test case to include commonmark example.

PR #22, opened 2022-10-26
test: link in heading

ChristianMurphy:
breaks expectation at
https://github.com/wooorm/markdown-rs/blob/b910e37bb387a62509a7fda26617e9870ed6d56f/src/to_html.rs#L197


wooorm:
This is about how definitions, setext headings, and lists interplay.
Here’s a reduced test case:

```rs
"[a]: b\n-\n-"
```

ChristianMurphy:
Also found a way to replicate the issue with only setext headings.
```markdown
a
-
--
a
-
```

which could potentially be related to #31, but shows up in a different way

PR #24, opened 2022-10-28
test: unordered list with GFM checkmark-like plain text

ChristianMurphy:
https://github.com/wooorm/markdown-rs/blob/5d5e94567e7a87a3cefdda410e3bd9e36e26eca5/src/to_mdast.rs#L1220-L1223


wooorm:
Found it!

PR #25, opened 2022-10-28
test: fuzz mdx parser

ChristianMurphy:
This adds `.ok()` to the `Result` to capture and ignore recoverable errors.


ChristianMurphy:
This may want to wait on #26, currently `panic!` gets triggered pretty quickly

fuzz/fuzz_targets/markdown.rs
```diff
@@ -9,7 +9,15 @@ fuzz_target!(|data: &[u8]| {
             s,
             &markdown::Options::gfm()
         );
+        let _ = markdown::to_html_with_options(
+            s,
+            &markdown::Options {
+                parse: markdown::ParseOptions::mdx(),
+                ..Default::default()
+            }
+        ).ok();
```
wooorm, line 18: This is pretty useless: MDX when going to HTML.
The latter one is useful: parsing MDX to an AST

fuzz/fuzz_targets/markdown.rs
```diff
@@ -9,7 +9,15 @@ fuzz_target!(|data: &[u8]| {
...
         let _ = markdown::to_mdast(s, &markdown::ParseOptions::default());
         let _ = markdown::to_mdast(s, &markdown::ParseOptions::gfm());
+        let _ = markdown::to_mdast(s, &markdown::ParseOptions::mdx()).ok();
```
wooorm, line 21: What does the `.ok()` add? I believe it’ll work fine without it: that it’s a result and doesn’t panic?

fuzz/fuzz_targets/markdown.rs
ChristianMurphy: removed

fuzz/fuzz_targets/markdown.rs
ChristianMurphy: My impression is/was that some errors are to be expected with MDX?
My goal is to capture/ignore known errors, and only look for unexpected `panic!`.

That understand could be incorrect though.
Is the goal with the `mdx` parser to never throw?

---

in the meantime removed `.ok`

ChristianMurphy: also related https://github.com/wooorm/markdown-rs/pull/26#issuecomment-1296280425

PR #26, opened 2022-10-28
test: handle invalid mdx with Result rather than panic

ChristianMurphy:
https://github.com/wooorm/markdown-rs/blob/5d5e94567e7a87a3cefdda410e3bd9e36e26eca5/src/tokenizer.rs#L802

This case goes to `unreachable!`
https://doc.rust-lang.org/std/macro.unreachable.html

> This will always [`panic!`](https://doc.rust-lang.org/core/macro.panic.html) because `unreachable!` is just a shorthand for `panic!` with a fixed, specific message.

which means `Result` is unable to recover.


wooorm:
This isn’t about the panic, which should be a result btw. It’s just that this `unreachable` should never be reached. Sort of like an assertion. And it is reached, which should be fixed somehow!

ChristianMurphy:
> This isn’t about the panic, which should be a result 

Meaning that it should not `panic!` or `Err`?
Or that `panic!` should automatically convert to `Err`?
Regarding the later. Rust has two types of exception handling.

> Rust groups errors into two major categories: recoverable and unrecoverable errors. For a recoverable error, such as a file not found error, we most likely just want to report the problem to the user and retry the operation. Unrecoverable errors are always symptoms of bugs, like trying to access a location beyond the end of an array, and so we want to immediately stop the program.
>
> Most languages don’t distinguish between these two kinds of errors and handle both in the same way, using mechanisms such as exceptions. Rust doesn’t have exceptions. Instead, it has the type `Result<T, E>` for recoverable errors and the `panic!` macro that stops execution when the program encounters an unrecoverable error. 

https://doc.rust-lang.org/book/ch09-00-error-handling.html

ChristianMurphy:
closed by #60

PR #28, opened 2022-11-09
Remove `extern crate`

Hocdoc:
Remove most of the `extern crate` lines.

Since Rust 2018 they are not needed anymore.


wooorm:
Looks like this touches something in the benchmarks that shouldn’t be touched?

wooorm:
Thank you! :)

PR #29, opened 2022-11-10
Fix `cargo fmt` errors

Hocdoc:
Fix the errors and warnings from the lint command: `cargo fmt --check && cargo clippy --examples --tests --benches`


wooorm (approved):

wooorm:
Thank you! :)

Issue #10, opened 2022-10-14
[feature-request] Make `Node` serializable to JSON


thiskevinwang:
👋 — awesome project here!

I'm not entirely sure if this is actually a helpful method to have, but a `.to_json()` method (or making `Node` serializable to JSON) on `Node` feels like a nice-to-have... 

I'm a rust noob, but I implemented said serialization functionality here in this repo:
https://github.com/thiskevinwang/markdown-rs-test

wooorm:
heyy 👋

Yeah, good idea to add serde-able annotations!

It would be important to match the JSON structure of https://github.com/syntax-tree/mdast though. Particularly to match exactly what https://github.com/syntax-tree/mdast-util-from-markdown exposes too.

kyle-mccarthy:
@wooorm I was looking into implementing Serialize for everything. Should this crate be parsing everything exactly has remark does?

wooorm:
a) What is *everything*?
b) Eh, yeah, at least adhering to `mdast`, where do you want to diverge?

kyle-mccarthy:
@wooorm I may be wrong, but I think that this implementation handles unicode characters differently than the javascript implementation in mdast. 

If you use the [link-reference test](https://github.com/syntax-tree/mdast-util-from-markdown/blob/main/test/fixtures/link-reference.md) and compare the outputs, the start and end points will all be off by 1 after encountering the fancy/smart apostrophe in the word "they're".

I assume this is because JS uses UTF-16 while Rust uses UTF-8, so Rust counts it as 2 where as in JS it's 1.

I have the serialization implemented, but the majority of the tests in `mdast-util-from-markdown` are failing.

wooorm:
You are correct. The positional info is different. This is indeed about how strings are present in the different languages.
It’s not perfect, but it makes some sense, as users do use this information to slice from the input document. In JS, its positional info is useable for that. In rust, these indices are usable.

But, why are you adding tests from `mdast-util-from-markdown`? That project is already implemented and tested here. This issue is about turning the Rust structures into JSON structures?

PR #30, opened 2022-11-21
feat: add `to_json` feature

kyle-mccarthy:
# Description

Add serialization for mdast and unist structs. Additionally, a `to_json` function was added that returns a `Result<serde_json::Value, Error>`.

# Closes

- #10 

# Tests

Since the goal was to match the JSON exposed by MDAST, I took the test input and output from that repo and added them as fixtures in this one. 

Some of the failing tests were caused by "smart quotes," so I converted those to their ASCII equivalent. 

However, there were also several failures not related to encoding. I've kept those, so this PR has tests that fail. I don't have enough background knowledge to determine if these are trivial differences or unintended divergences.


wooorm:
Hey, Awesome! Thanks for working on this!

Couple Qs:

a) Why is a `to_json` needed, that is exactly the same as `to_mdast`, with the exception that it does `serde_json::to_value(&node)`?
b) Have you thought about a list of test cases in `tests/to_json.rs` (or even reading the folder), and looping through them, to generate test cases?
c) Is there a reason you are not using the `derive` macros? https://serde.rs/derive.html. It seems to me like that would involve a lot less code!
d) If keeping it as code, have you thougt about inlining the `ser` modules with their corresponding `mdast`, `unist` files? I’d think that would make it easier to maintain them if they are adjacent to each other.

> I've kept those, so this PR has tests that fail.

Which tests fail? 🤔 CI passes?

kyle-mccarthy:
Hey thanks for looking this over. 

1. The reason that I made `to_json` was so I could return specific error types rather than just return a string value for an error. Ideally, when returning a `Result<T, E>` the `E` type would implement `std::error::Error`. I understand this goes against how the library handles errors, but I wanted to provide the option. I could definitely just call `to_mdast` from `to_json` if you would prefer.
2. I actually did have a node script that would generate the JSON test cases based on the directory contents. I didn't include this since I assumed additional test cases added would be more of a one off thing rather than being added in bulk. Also I made them as individual functions rather than a single function that reads the directories and performs the tests so you can run specific cases one off like `cargo test --test to_json test_list -- --exact`.
3. Yeah this would have been the easiest path, however, I don't think that there is an easy way to generate the `type` tag in the payload. Especially for the case like `mdxJsxAttributeValueExpression`. 
4. The reason that I made them separate is due to how rust handles modules. Hypothetically, if the fields on the Point and Position weren't public, you wouldn't be able to read them from the `mdast::ser` module. However, you would be able to read them from `unist::ser` since `Point` and `Position` are located in the parent module.

Regarding the failing tests missing from CI. I made the `serde` feature optional so the corresponding tests won't be run unless the feature is enable.d I think that CI will need to be updated to run `cargo test --all-features`.

wooorm:
Re 1): I can see some value in providing error objects (especially for line/col/reason info), but I’d want to discuss how to do that separately, and then apply it everywhere

Re 2) I saw the script, and I indeed don’t think it’s useful to include it. I also see some value in having separate test cases so you can indeed do `--test ...`. But I worry about a) the amount of code needed to run all of this, which is hard to maintain and update, b) the boring task that maintainers need to perform to add test cases, c) contributors adding files but forgetting to include a test case.
Perhaps you can generate a test file with separate test cases from the fixtures, similar to how the CommonMark test case is built? https://github.com/wooorm/markdown-rs/blob/af962122d2e063aabece64296bce004648802dfc/build.rs#L10

Re 3) So you are saying there is no way to include a field in the JSON, if it isn’t in the Rust struct either? If that’s the case, we can add a `kind` field to the rust structs, and rename it to `type` with Serde?

Re 4) I don’t get this reason. You are correct that at some point `unist` might not be exposed from here. But then it would be exposed from a different `unist` crate.

Re errors: these are interesting, seems like there are some things we can fix in this project :)
Idea: if/when you implement a script like the CM stuff in `build.rs` for the tests, generate a skipped test case if actual != expected, so we can solve them later?

kyle-mccarthy:
@wooorm I made an alternative impl to this on a different branch [feat/json](https://github.com/kyle-mccarthy/markdown-rs/blob/feat/json/src/mdast.rs). I was looking through serde's issues and realized there was an undocumented feature that let you tag struct's directly. Let me know what you think and which one you prefer!

wooorm:
Oh I like that a lot! Do you have any reservations!

kyle-mccarthy:
None at all. I'll open up a PR and close this one in the next day or so with that impl. I was going to keep the tests as-is but ignore the ones that fail from the js implementation (the ones that fail because of the difference in position).

PR #35, opened 2022-12-13
build(cargo): allow to control reqwest tls backend

kwonoj:
This PR allows to control TLS backend for the dependency `reqwest`, while makes default (native-tls) as-is. Only if someone need to explicitly control they can disable default, then opt into specific.

`reqwest` is build-dependencies, so may seems it doesn't cause any problem: however, due to cargo creates lockfile & compiles it can actually brings some interesting challenges. 

```
// Cargo.toml: Application

[dependencies]
LibA = *

// Cargo.toml: LibA
[dependencies]
reqwest = {.., features = ["rustls-tls"]}
markdown = *

// Cargo.toml: markdown

[build-dependencies]
reqwest = * // default, native-tls is enabled
```

Now building `Application` generates a lockfile for it, and cargo _merges_ every feature for the dependencies into single like 

```
reqwest = {
...
features = ["rustls-tls", "native-tls"]
}
```

regardless of type of dependencies. There are target platforms require specific TLS backend only, now building `Application` fails on those platform.

I tried to make this change non-invasive as much. Default works exactly same as current, and only explicit opt-in would change its behavior. Also added cases for 1. for disabling tls entirely, attempt to use `http` with warning 2. specifying both tls fails build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,13 @@ homepage = "https://github.com/wooorm/markdown-rs"
 repository = "https://github.com/wooorm/markdown-rs"
 license = "MIT"
 keywords = ["commonmark", "markdown", "parse", "render", "tokenize"]
-categories = ["compilers", "encoding", "parser-implementations", "parsing", "text-processing"]
+categories = [
+  "compilers",
+  "encoding",
+  "parser-implementations",
+  "parsing",
+  "text-processing",
+]
 include = ["src/", "license"]
 
 [[bench]]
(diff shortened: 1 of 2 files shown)


wooorm:
Did you see https://github.com/wooorm/markdown-rs/issues/34?
I mistook what “build” means in Rust, so the plan is to move it to dev dependencies.

Would doing so that also solve this issue?

wooorm:
From what I see dev-dependencies of dependencies are not in cargo.locks, so that should solve it?

kwonoj:
If that's the case, yes, I think so.

kwonoj:
Just to clarify I'm fine either cases to move away dep from build, or trying this way - please feel free to close this one if needed. I also think moving dep away would be more preferred way.

wooorm:
Fixed in https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.5!

PR #37, opened 2022-12-19
Add support for serializing trees as JSON

kyle-mccarthy:
# Description

This is a continuance of PR #30 that leverages serde's macros instead of explicit impls for serialize.

# Closes
- #10 

# Tests
Test have been taken from the `mdast-util-from-markdown` JS package. Tests that were failing have been `#[ignore]`d.


src/lib.rs
```diff
@@ -146,3 +146,19 @@ pub fn to_mdast(value: &str, options: &ParseOptions) -> Result<mdast::Node, Stri
...
+/// returned if the syntax tree cannot be serialized to JSON.
+#[cfg(feature = "json")]
+pub fn to_json(value: &str, options: &ParseOptions) -> Result<serde_json::Value, String> {
+    let (events, parse_state) = parser::parse(value, options)?;
+    let node = to_mdast::compile(&events, parse_state.bytes)?;
+
+    let value = serde_json::to_value(&node).map_err(|_| "Failed to serialize mdast to json")?;
+
+    Ok(value)
+}
```
wooorm, line 164: I don’t really think this is needed: I think we should keep the core API small, and document how people can turn any Node into a JSON string.

src/mdast.rs
```diff
@@ -428,14 +443,19 @@ impl Node {
...
+)]
 pub enum AttributeContent {
     /// JSX expression.
     ///
     /// ```markdown
     /// > | <a {...b} />
     ///        ^^^^^^
     /// ```
-    Expression(String, Vec<Stop>),
+    Expression { value: String, stops: Vec<Stop> },
```
wooorm, line 458: Should this reuse `AttributeValueExpression`, perhaps then renamed something else?

src/mdast.rs
```diff
@@ -444,23 +464,45 @@ pub enum AttributeContent {
...
     /// > | <a b={c} />
     ///          ^^^
     /// ```
-    Expression(String, Vec<Stop>),
+    #[cfg_attr(feature = "json", serde(rename = "mdxJsxAttributeValueExpression"))]
+    Expression {
+        value: String,
+        #[cfg_attr(feature = "json", serde(skip))]
+        stops: Vec<Stop>,
+    },
```
wooorm, line 498: Same as above: should this also reuse a struct?

wooorm (commented):
Nice! I think this is almost there :) Thanks for your continued work.

One more thing, is that I still think it’s good to have a build script, to take test fixtures, and generate test cases, like what you had inline before, as mentioned here: https://github.com/wooorm/markdown-rs/pull/30#issuecomment-1331891073 (point 2), in the file https://github.com/wooorm/markdown-rs/blob/main/generate/src/main.rs

src/lib.rs
kyle-mccarthy: Sure this could be removed. I mainly just added it for convenience.

src/mdast.rs
kyle-mccarthy: hmm possibly... is there a difference between an `mdxJsxExpressionAttribute` and `mdxJsxAttributeValueExpression`?

kyle-mccarthy:
> One more thing, is that I still think it’s good to have a build script, to take test fixtures, and generate test cases, like what you had inline before...

I am totally open to doing this; however, I am not sure how to handle:
- the files that I've manually had to change to account for utf-8
- the failing test cases that have been ignored using the ignore macro

wooorm:
> the files that I've manually had to change to account for utf-8

The input is UTF-8, so I don’t understand why you need to account for this?

> the failing test cases that have been ignored using the ignore macro

I’d go about this as follows: you could use this project itself in that build script to check if `actual` is `expected`, in which case you generate a normal test. Otherwise, you generate a skipped test: https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested

src/mdast.rs
wooorm: Well, they all have the same fields, so those are the same.
The only difference is where they occur, and I think the name of the enum itself, already specifies it well!

kyle-mccarthy:
Hey so I have a couple of concerns:

- do you feel like the build script is a good place for this? It gets run by rust-analyzer by default so downstream projects using the crate will have the tests get automatically downloaded and run when their LSP starts.
- will having the build script automatically mark failing tests as ignored will hide regression? e.x. a change is made, the build script creates the test, marks a test that was previously okay as ignored when it starts to fail.
- I manually fixed several inputs of the failing tests to account for the different positions in rust vs javascript, I don't know if this can be automatically done.

Alternatively,I could just remove the tests?

wooorm:
a) this is no longer the case since https://github.com/wooorm/markdown-rs/commits/main
b) it doesn’t run all the time anymore and the output is checked in so visible
c) you could remove positional info before testing
d) I am also okay with removing most tests. I don’t believe the core of the issue you are solving here is about matching the exact output of `mdast-util-from-markdown`

kyle-mccarthy:
@wooorm ah nice I didn't realize that the test situation changed! that said, I went with option d since I agree that the tests I added were overkill for this feature.

wooorm (approved):

wooorm:
Thanks for your hard work on this feature!!

Issue #33, opened 2022-12-09
Future-proof Options before 1.0


Ekleog:
Hey! I was having a look at markdown, and noticed that the latest 1.0 alpha has Options that are structs.

I'm just coming to say, if you make it into a builder then it'll allow you to add options with minor releases, whereas with a struct like right now any option addition would require a full major release.

(This is also something I'd probably need: I'll probably soon-ish want to compile inline-only markdown to "markdown-html" (html that retains the markdown tags), and if I want any chance to upstream it I'll need the possibility of adding both parse and compile option. That said I'm not even using this crate yet, so this may totally never happen)

For the same reason you may want to make the `Node` enum non-exhaustive, though I'm less sure about this one. But the more stuff you can make non-exhaustive / made out of builder patterns with private members without hindering user experience, the less likely you'll need major updates in the future :sweat_smile:

wooorm:
As I understand, `#[non_exhaustive]` can solve all these issues, right? Forcing folks to spread the defaults in?
And a builder pattern is another, different, way to solve it, for options?

Re “inline only”, I’m not sure I think that’s a good feature to have. Depends…

Ekleog:
##### For options (structs)

So I've seen `#[non_exhaustive]` used mostly on enums up to now. I guess non-exhaustive structs would require code like:
```rust
let mut options: Options = Default::default();
options.foo = true;
options.bar = true;
do_something(options);
```
which is less pleasant than (builder-pattern equivalent):
```rust
do_something(Options::builder()
    .foo(true)
    .bar(true)
    .build());
```

The builder pattern is basically adding methods to `Options` to enable the second syntax, whereas afaiu non-exhaustive structs can currently not be constructed even by completing them with `..Default::default()`.

(This being said, both options do avoid the issue that adding one field is a breaking change; it's just that one is easier to handle for the library user)

Note also that there are crates like https://crates.io/crates/derive_builder to reduce the boilerplate of writing builder patterns :)

#### For enums

Now, I also mentioned making the Node enum non-exhaustive; this is mostly to avoid that people do a `match` on it: without it, adding one variant to `Node` would be a breaking change.

(As for the inline-only, I'm not opening this issue to discuss it yet, we can chat about it if/when I actually get to it; it was just to showcase one possible example of a reason why being able to add options could be useful :))

wooorm:
[...]

PR #40, opened 2023-01-02
Use a builder pattern for public structs, for future-proofing

Ekleog:
See discussion at https://github.com/wooorm/markdown-rs/issues/33


src/mdast.rs
```diff
@@ -447,6 +450,7 @@ pub enum AttributeContent {
 
 /// MDX: attribute value.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
 pub enum AttributeValue {
```
wooorm, line 454: I think these are supposed to be exhaustive?

wooorm (commented):
Thanks for working on this!

* Why is the `pub(crate)` stuff needed?
* I think the helper fns for `mdx`, `gfm`, and such, should not generate a new options struct, but instead operate on an existing one, that way, it could be used as `::default().gfm(true).mdx(true)`, to turn on both
* should we still expose the structs too, and then also expose `*Builder` for each? Why not only expose builders, under the original name? As an example, why `CompileOptionsBuilder::default().allow_dangerous_html(true).build()`, instead of something like `compile_options().allow_dangerous_html(true).build()`
* Can you rebase? :)

src/mdast.rs
Ekleog: For AttributeValue and AttributeContent, I'll trust you, as I don't even know what they're about as I never used JSX markdown ^^'

For Node, I'm pretty sure keeping it non-exhaustive would be better: it seems likely to me that some future evolution will require adding node types given the length of the enum (eg. supporting some new variant of markdown?). And someone who already filled match arms for each of the variants can probably also fill in a default match arm for any future evolution :sweat_smile: 

For ReferenceKind, it's kind of the same train of thought, but I'm less sure about myself.

Does my thinking make sense? If so, which ones should I turn exhaustive back?

Ekleog:
>     * Why is the `pub(crate)` stuff needed?

It allows other modules inside the same crate to access the value of these properties. The option structs are not defined in the same module as the place where they're used from, so this is required to allow access to them from there.

>     * I think the helper fns for `mdx`, `gfm`, and such, should not generate a new options struct, but instead operate on an existing one, that way, it could be used as `::default().gfm(true).mdx(true)`, to turn on both
>
>     * should we still expose the structs too, and then also expose `*Builder` for each? Why not only expose builders, under the original name? As an example, why `CompileOptionsBuilder::default().allow_dangerous_html(true).build()`, instead of something like `compile_options().allow_dangerous_html(true).build()`

I see what you mean, and had the same thought, but I couldn't find a crate to easily do that without manually writing all the code. (to be precise, safe-derive-builder does do it, but has 1k downloads and the last update is from 5 yrs ago, plus exponential build times, considering the number of parameters, sounds bad).

Looking at it again, literally while writing this comment I just came across derive-setter, that could probably be exactly what you're looking for.

However, to be quite honest, fixing all the tests to use the new API definitely exhausted my will to contribute this for now, and I won't actually do the work of porting them to a new API again, as that'd be almost as long as starting from scratch again. Sorry about that!

But yes, porting this PR to derive-setter would probably be the best. I still think that this PR is already an improvement, and will make any future switch to derive-setter easier; but I'll let you judge on that.

>     * Can you rebase? :)

Well, knowing the above, do you think it'd be useful that I rebase? If yes, I'll be happy to, to push this PR past the finish line, but I can't promise more

Issue #27, opened 2022-11-01
`markdown::to_hast` support


pd4d10:
Would it be supported in the future plan?

wooorm:
Feel free to take the code for that now from here: https://github.com/wooorm/mdxjs-rs/blob/main/src/mdast_util_to_hast.rs.

I don’t really know how to best expose those parts. And you will likely need more parts (hast-util-to-html). I don’t think they make a lot of sense here. Depends on how people want to use these things!

kwonoj:
Just fyi, I'm also in need of similar api exposure from mdxjs-rs, though I'm also not sure final shape of surface yet (under experimenting now).

@wooorm Is this something acceptable in upstream mdxjs-rs PR to expose few interfaces as public? or the recommendation is copy-code-as needed?

wooorm:
Can you tell me more about what you are working on?

My previous comment still stands: you will likely need more things.
I’d probably want to expose several unist, mdast, and hast utilities, each in a different monorepo for their AST?

Keeping it hidden for now means that some of the AST can still be improved. I’d particularly first like to:

- Use traits on each node instead of specifying things globally (e.g. [for context see this](https://github.com/wooorm/markdown-rs/blob/af962122d2e063aabece64296bce004648802dfc/src/mdast.rs#L212)), that would allow unist utilities (such as [this](https://github.com/wooorm/mdxjs-rs/blob/8cba71bcf926057ae4bdc5b67de69010a5621301/src/mdast_util_to_hast.rs#L1197-L1222)) to work on both mdast and hast
- Make AST nodes serializable in a good way: https://github.com/wooorm/markdown-rs/pull/30

After those two, I think it should be a stable API that can be exposed either from the crates currently or later from separate packages

kwonoj:
Sure, the current work I'm trying to do is integrate mdx support in turbopack (https://github.com/vercel/turbo).

I'm still in the process of figuring out what's necessary piece, so I may wrong, but so far I came to conclude it requires 2-stage process for `parse(input, option) -> AST` / `transform(AST, option)` as separate. This requires some internal access to existing mdxjs-rs to split things up and create separate api surfaces.

Again, please note this is not a final form and reason I mentioned `though I'm also not sure final shape of surface yet (under experimenting now).`. I would like to get more clarity before making any actual upstream effort.

wooorm:
Cool, interesting!

> so far I came to conclude it requires 2-stage

[...]

PR #46, opened 2023-02-04
implement pre-order traversal of node

remodietlicher:
Hello! I'm playing around with your nice library to learn a bit of Rust. I found myself filtering and mapping the AST, for example to extract only a specific type of Node. After implementing my own `filter` function I noticed that the Iterator trait actually provides just that, all that's needed is the `next` function.

So here I implement a pre-order traversal of the Node-tree to create an Iterator over the AST. I show how it can be used on the `to_string` implementation.

Since I'm new to Rust, I was wondering if this sounds useful to you and if there is anything I could improve.

### DIFF
--- a/src/mdast.rs
+++ b/src/mdast.rs
@@ -6,6 +6,7 @@ use crate::unist::Position;
 use alloc::{
     fmt,
     string::{String, ToString},
+    vec,
     vec::Vec,
 };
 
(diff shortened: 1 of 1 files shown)


wooorm:
Hey! Thanks for your patience, cool work!

This seems a bit similar to my preorder traversel utility here: https://github.com/wooorm/mdxjs-rs/blob/8cba71bcf926057ae4bdc5b67de69010a5621301/src/mdast_util_to_hast.rs#L1198-L1222. (I also had a mutable version there as well: https://github.com/wooorm/mdxjs-rs/commit/f6b705f2fe7c47ff704aeca563f25c5fd5b81850#diff-d1bf57e8534841c5ee7eda0ab2bcf89b943719c21ea89a48bf96e25a114e2f3f)

I’ve been meaning to turn the nodes into traits (see https://github.com/wooorm/markdown-rs/issues/27#issuecomment-1340467635). That way, we can have such utilities working on mdast *and* hast.
Perhaps that’s something you might be interested in working on? :)

wooorm:
Closing; I think I would be open to a a `unist_util_visit` fork as a subcrate here!

PR #48, opened 2023-02-06
Move `log` to optional dependencies

pinkforest:
Log is unnecessary for production code so this can be turned into optional whilst debugging.

This saves code sizes and dependencies on pre-1.0 crates.

fwiw - there is also tracing may be worth consideration in future

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ default = ["json"]
 json = ["dep:serde", "dep:serde_json"]
 
 [dependencies]
-log = "0.4"
+log = { version = "0.4", optional = true }
 unicode-id = { version = "0.3", features = ["no_std"] }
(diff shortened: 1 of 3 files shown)


wooorm (commented):
Can you document this feature? Perhaps around https://github.com/wooorm/markdown-rs#install.

You can also make a short `Debug` section. Around https://github.com/wooorm/markdown-rs#version. For example like https://github.com/micromark/micromark#size--debug.

wooorm (approved):

Cargo.toml
```diff
@@ -22,7 +22,7 @@ default = ["json"]
 json = ["dep:serde", "dep:serde_json"]
 
 [dependencies]
-log = "0.4"
+log = { version = "0.4", optional = true }
```
wooorm, line 25: Shouldn’t a feature be added for this?

PR #49, opened 2023-02-06
Remove `json` from default features

### LINKED ISSUE #47: Should json be default feature ?
$ cargo update
```
    Updating crates.io index
      Adding itoa v1.0.5
    Updating markdown v1.0.0-alpha.5 -> v1.0.0-alpha.6
      Adding ryu v1.0.12
      Adding serde v1.0.152
      Adding serde_json v1.0.92
```

I am using mdast just to parse it and then process the mdast structures, e.g. I don't need json or any serialization out.

Would there be openness to having it as optional considering there are different ways to serialize out ?

Thanks for the great work btw ! :purple_heart: 

Most of the ecosystem has serde as optional feature

pinkforest:
Fixes #47 

json serialization is supplementary feature and ecosystem would assume this would be optional faeture on top of parsing

Aligns to ecosystem re: default primary feature set whilst reducing codesizes & maintenance overhead on-by-default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ path = "benches/bench.rs"
 harness = false
 
 [features]
-default = ["json"]
+default = []
 json = ["dep:serde", "dep:serde_json"]
 
 [dependencies]


wooorm (approved):

wooorm:
Thanks!

PR #51, opened 2023-02-13
Fix AST node debug representation to pretty print

ghost:
The current implementation breaks `{:#?}` formatting and instead displays it as what `{:?}` would say. This change makes it so we use the correct formatter, meaning it will pretty print when asked.


wooorm (approved):

wooorm:
Thanks!

wooorm:
[Released](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.8), thanks!

PR #60, opened 2023-04-20
Fix #26 EOF handling in es_whitespace_inside

mickvangelderen:
Since @ChristianMurphy has gone through the trouble of exposing some issues with the parsing, I thought I would try to fix one https://github.com/wooorm/markdown-rs/pull/26. 

While this prevents the program from reaching the `unreachable!` statement
when parsing `"<"`, I do not know if the solution is correct. I do not
completely understand how the parser works and why this issue occurred
and so the solution direction presented in this commit might not be the
best.

### DIFF
--- a/tests/fuzz.rs
+++ b/tests/fuzz.rs
@@ -117,5 +117,17 @@ fn fuzz() -> Result<(), String> {
         "11: gfm task list items followed by eols (GH-24)"
     );
 
+    assert_eq!(
+        markdown::to_html_with_options(
+            "<",
+            &markdown::Options {
+                parse: markdown::ParseOptions::mdx(),
+                ..Default::default()
+            }
+        ),
+        Ok("<p>&lt;</p>".to_string()),
+        "10: mdx: handle invalid mdx without panic"
+    );
+
     Ok(())
 }
--- a/tests/fuzz.rs
+++ b/tests/fuzz.rs
@@ -126,7 +126,7 @@ fn fuzz() -> Result<(), String> {
             }
         ),
         Ok("<p>&lt;</p>".to_string()),
-        "10: mdx: handle invalid mdx without panic"
+        "12: mdx: handle invalid mdx without panic (GH-26)"
     );
 
     Ok(())
--- a/src/construct/partial_mdx_jsx.rs
+++ b/src/construct/partial_mdx_jsx.rs
@@ -1070,17 +1070,19 @@ pub fn es_whitespace_inside(tokenizer: &mut Tokenizer) -> State {
             tokenizer.consume();
             State::Next(StateName::MdxJsxEsWhitespaceInside)
         }
-        _ => {
+        Some(_)
             if kind_after_index(tokenizer.parse_state.bytes, tokenizer.point.index)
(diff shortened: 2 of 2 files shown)


wooorm:
There are several places that use this code.
I think `kind_after_index` should just yield `characterkind::Whitespace` on EOF (index == bytes.length), and that should solve all of those potential issues?

wooorm:
A similar case seems to be: https://github.com/wooorm/markdown-rs/blob/b3a921c761309ae00a51fe348d8a43adbc54b518/src/construct/gfm_autolink_literal.rs#L368-L370.

wooorm:
I recommend simplifying your check:

```diff
        // Handle EOF.
-        _ if tokenizer.point.index == tokenizer.parse_state.bytes.len() => {
-            State::Nok
-        }
+        None => State::Nok
```

mickvangelderen:
> There are several places that use this code.
> I think `kind_after_index` should just yield `characterkind::Whitespace` on EOF (index == bytes.length), and that should solve all of those potential issues?

I believe it already does return `CharacterKind::Whitespace`. That in turn causes the `consume` call which tries to read past the end of the byte slice triggering the `unreachable!` statement. Perhaps the `kind_after_index` should return a new variant for EOF and all of the invocations need to be updated to handle the possibility of EOF?

wooorm:
Ah good.

Then only my last comment (https://github.com/wooorm/markdown-rs/pull/60#issuecomment-1516065756) is needed

mickvangelderen:
> I recommend simplifying your check:
> 
> ```diff
>         // Handle EOF.
> -        _ if tokenizer.point.index == tokenizer.parse_state.bytes.len() => {
> -            State::Nok
> -        }
> +        None => State::Nok
> ```

:+1:

mickvangelderen:
@wooorm let me know if you are okay with the refactor or if you would like me to revert it.

wooorm:
I’m not sure why you now went with a big refactor and a new function, instead of the one line change of `None => State::Nok`?

What is the benefit of all this changed code over such a small change?

mickvangelderen:
@wooorm I think the PR is ready for review and merge with your approval

ChristianMurphy (approved):
Thanks @mickvangelderen!

tests/fuzz.rs
```diff
@@ -117,5 +117,17 @@ fn fuzz() -> Result<(), String> {
...
+    assert_eq!(
+        markdown::to_html_with_options(
+            "<",
+            &markdown::Options {
+                parse: markdown::ParseOptions::mdx(),
+                ..Default::default()
+            }
+        ),
+        Ok("<p>&lt;</p>".to_string()),
+        "10: mdx: handle invalid mdx without panic"
```
wooorm, line 129:
```suggestion
        "12: mdx: handle invalid mdx without panic (GH-26)"
```

wooorm (approved):

wooorm:
[Released](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.8), thanks!

PR #70, opened 2023-05-23
Make `Options` (De)serialisable with serde

byhemechi:
I've added basic serde attributes to the `Config` struct and its dependencies, only within the serde feature.

Please have a brief check that i've not missed any renames so that we stay in sync with micromark

Feel free to be rude if i've done anything stupid

All existing tests pass though I haven't written any new ones yet


byhemechi:
please don't merge this yet, i'm yet to set up the defaults properly

byhemechi:
The check fail is on a file outside the scope of this PR so i'm not going to fix it here

wooorm:
yeah don’t worry about that, that was my commit just now, will fix soon!

wooorm:
Thanks! Code looks good, I do wonder if there should be some tests? And, should be documented in the features docs: https://docs.rs/markdown/1.0.0-alpha.9/markdown/#features

byhemechi:
Just writing some tests now, is it acceptable to pull in serde_json as a dev dependency for the feature?

byhemechi:
I've adjusted the docs, will work on tests very soon. Are you ok with pulling in serde_json as a dev dependency for said tests?

wooorm:
Sure!

byhemechi:
Alright, doing test now. To be entirely honest i don't know if these test actually achieve anything as all they are really doing is testing `serde` behaviour, which is already in their tests and is obviously covered by semver

wooorm:
I guess you can check if it works, and that there is a specific string being generated?

byhemechi:
yeah, that's what i'm ending up doing :)

wooorm:
last I remember you were still working on this, right? Or is it done?

byhemechi:
Sorry, got busy with life and didn't quite finish it :)

I've written some of the checks but once again they dont really increase coverage, serde has tests on the macros and the structs themselves already have tests

PR #73, opened 2023-07-09
Fix parsing of nested links

### LINKED ISSUE #50: Internal panic with nested links
The strings
`[]()`
`]()`
cause panics when passed to `to_mdast`!

I'm surprised this didn't get detected by fuzzing - you might want to double check the fuzzer

sheremetyev:
Enter event for LabelText should be inserted before existing events at the index in order to have correct nesting. Otherwise nested elements could have Enter event first and that would result in incorrect nesting in the tree when converting to AST.

Fixes #50


wooorm:
Thank you, released! :)

PR #74, opened 2023-07-11
Fix serde

### LINKED ISSUE #72: MDAST: serde Serialization/Deserialization not working
The serialization of the mdast fails to deserialize because of a duplicate field type

```rust
#[test]
fn test_serde() {
    let markdown = "This is a **test**";
    let tree =
    markdown::to_mdast(&markdown, &markdown::ParseOptions::default()).unwrap();
    let json = serde_json::to_string(&tree).unwrap();
    let tree2: Node = serde_json::from_str(&json).unwrap();
    assert!(tree == tree2);
}
```
I get the following error from serde
```
Error("duplicate field `type`", line: 1, column: 21)
```

This is the generated json. It has duplicate type fields everywhere. 
```json
{
   "type":"Root",
   "type":"root",
   "children":[
      {
         "type":"Code",
         "type":"code",
         "value":"This is a **test**",
         "position":{
            "start":{
               "line":2,
               "column":1,
               "offset":1
            },
            "end":{
               "line":3,
               "column":5,
               "offset":28
            }
         },
         "lang":null,
         "meta":null
      }
   ],
   "position":{
      "start":{
         "line":1,
         "column":1,
         "offset":0
      },
      "end":{
         "line":3,
         "column":5,
         "offset":28
      }
   }
}
```

Looking at the AST `enum` and related `struct`s they all have `serde(tag, rename)` macro configs, which seem to be conflicting. 
```rust

#[derive(Clone, Eq, PartialEq)]
#[cfg_attr(
    feature = "serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(tag = "type", rename = "type")
)]
pub enum Node {
    // Document:
    /// Root.
    Root(Root),
//..
}

#[cfg_attr(
    feature = "serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(tag = "type", rename = "root")
)]
pub struct Root {
    // Parent.
    /// Content model.
    pub children: Vec<Node>,
    /// Positional info.
    pub position: Option<Position>,
}

```

sshwy:
This pr is trying to fix #72.

### DIFF
--- a/src/mdast.rs
+++ b/src/mdast.rs
@@ -86,95 +86,129 @@ pub enum AlignKind {
 #[cfg_attr(
     feature = "serde",
     derive(serde::Serialize, serde::Deserialize),
-    serde(tag = "type", rename = "type")
+    serde(tag = "type")
 )]
 pub enum Node {
     // Document:
     /// Root.
+    #[serde(rename = "root")]
     Root(Root),
 
     // Container:
     /// Block quote.
+    #[serde(rename = "blockquote")]
     BlockQuote(BlockQuote),
     /// Footnote definition.
+    #[serde(rename = "footnoteDefinition")]
     FootnoteDefinition(FootnoteDefinition),
     /// MDX: JSX element (container).
+    #[serde(rename = "mdxJsxFlowElement")]
     MdxJsxFlowElement(MdxJsxFlowElement),
     /// List.
+    #[serde(rename = "list")]
     List(List),
 
     // Frontmatter:
     /// MDX.js ESM.
+    #[serde(rename = "mdxjsEsm")]
     MdxjsEsm(MdxjsEsm),
     /// Toml.
+    #[serde(rename = "toml")]
     Toml(Toml),
     /// Yaml.
+    #[serde(rename = "yaml")]
     Yaml(Yaml),
 
     // Phrasing:
     /// Break.
+    #[serde(rename = "break")]
     Break(Break),
     /// Code (phrasing).
+    #[serde(rename = "inlineCode")]
     InlineCode(InlineCode),
     /// Math (phrasing).
+    #[serde(rename = "inlineMath")]
     InlineMath(InlineMath),
     /// Delete.
+    #[serde(rename = "delete")]
     Delete(Delete),
     /// Emphasis.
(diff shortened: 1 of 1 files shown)


wooorm:
Can you add a a test that breaks before your change, and works now? Thanks.

sshwy:
> Can you add a a test that breaks before your change, and works now? Thanks.

I've added a simple test for it. I know it is not sufficient to prove that the result of serialization is consisitent with [mdast](https://github.com/syntax-tree/mdast)'s type definition. However I'm getting a little busy these days, so it may take a long time for me to write a more thorough test if you urge :).

Cargo.toml
```diff
@@ -38,6 +38,7 @@ swc_core = { version = "0.78.0", features = [
   "ecma_parser",
   "common",
 ] }
+serde_json = "1.0"
```
wooorm, line 41: Can you move this one up above swc, so it is sorted?

tests/mdast_serde.rs
```diff
@@ -0,0 +1,10 @@
+use markdown::mdast::Node;
+
+#[cfg_attr(feature = "serde", test)]
+fn test_serde() {
+    let markdown = "This is a **test**\n\n> quote content\n> continue";
+    let tree = markdown::to_mdast(&markdown, &markdown::ParseOptions::default()).unwrap();
+    let json = serde_json::to_string(&tree).unwrap();
+    let tree2: Node = serde_json::from_str(&json).unwrap();
+    assert!(tree == tree2);
+}
```
wooorm, line 10: Thanks. Can you add one more, simple, test, to check the *actual* JSON?

Because JSON *allows* duplicate fields. It’s just that currently, perhaps, serde doesn’t parse that. But that might change in the future.

I’d like to know that a *string* of JSON does not include duplicate fields, instead of whether Serde can parse its own output! (I’d assume it always would do that?)

wooorm (commented):
2 small nits

tests/mdast_serde.rs
sshwy: done

wooorm:
Tests fail, I think this is because you’re now testing serde, but it’s an *optional* dependency, and should be a *required* dev-dependency?

PR #77, opened 2023-08-05
Add support for enabling GFM tasklist checkedness

cellularnetwork:
This allows you to to enable whether or not html checkbox inputs are checkable, i.e. do not have the `disable=""` attribute.

Apologies if the terminology is confusing, I felt it was necessary to make it immediately obvious that this doesn't enable/disable checkboxes but instead enables/disables their ability to be checked.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -831,6 +831,38 @@ pub struct CompileOptions {
     /// ```
     pub gfm_footnote_clobber_prefix: Option<String>,
 
+    /// Whether or not GFM task list html `<input>` items are enabled.
+    ///
+    /// This determines whether or not the user of the browser is able
+    /// to click and toggle generated checkbox items. The default is false.
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// use markdown::{to_html_with_options, CompileOptions, Options, ParseOptions};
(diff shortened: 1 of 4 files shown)


wooorm:
And then? How would you use this? Uncontrolled inputs don't do anything in HTML?

wooorm:
Wouldnt you want to safe that somewhere? A server or local storage?

wooorm:
> no, i don't see a need to, it's just a cute thing for people to use that will help them follow the guide while they are on the page.

I worry about this. A nice cute thing is fun and all, of course.
But these inputs are not labelled, they are not in a form, I worry about how screenreaders for example handle them.

wooorm:
ok, not against trying this out!
Looks like CI is failing? See these for help:
- https://github.com/wooorm/markdown-rs/blob/main/.github/contribute.md#submitting-a-pull-request
- https://github.com/wooorm/markdown-rs#test

wooorm:
If you’re interested in fixing it, in this PR, or even separately, then please be my guest: I appreciate it! Whatever works for you :)

wooorm (approved):

wooorm:
Thanks!

wooorm:
Released! :)

PR #82, opened 2023-09-23
Fix `tarpaulin` out format

FirelightFlagboy:
This change will fix the failure on https://github.com/wooorm/markdown-rs/actions/runs/6252150859/job/16974753952


wooorm:
Ah, yeah I was wondering what was bugging out with the coverage, I assumed it was maybe a broken release of tarpaulin that would solve itself. Thanks for this PR!

wooorm:
Thank you!

PR #88, opened 2023-11-12
issue 84: GFM strikethrough causes nested attention sequences to be considered just text data

rawilder:
The original issue I was trying to solve was that a markdown line like this `~~foo __*a*__~~` was not correctly parsed because tilde was not being considered punctuation.

After changing that, the interplay between emphasis and strikethrough started failing because to implement how Github flavors it, we must consider tilde _not_ punctuation when deciding if the sequence is the close.  The interplay example is `e ***~~xxx~~***yyy`

I've added a specific exception for tilde with a comment explaining why it is there.

### DIFF
--- a/src/construct/attention.rs
+++ b/src/construct/attention.rs
@@ -81,10 +81,7 @@ use crate::resolve::Name as ResolveName;
 use crate::state::{Name as StateName, State};
 use crate::subtokenize::Subresult;
 use crate::tokenizer::Tokenizer;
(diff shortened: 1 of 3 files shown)


wooorm:
Thank you Adam for your work and your patience, that was interesting to solve.

I solved it more similar to how it is done in micromark.
I also noticed another bug present here: some ASCII punctuation characters are not Unicode punctuation.

PR #103, opened 2024-01-29
Mark mdast lists as spread if any children are spread

begleynk:
Firstly, thanks for creating a great library! 🚀 

I ran into what I think is a bug in how `List` nodes are marked as `spread`. The [docs](https://docs.rs/markdown/1.0.0-alpha.16/markdown/mdast/struct.List.html#structfield.spread) say this:

> One or more of its children are separated with a blank line from its siblings (when true), or not (when false).

In my testing this didn't seem to be the case. I'd see a spread `ListItem`, but the parent `List` was still marked as _not_ spread.

I looked into the code, added a failing test, and it seemed like `list_loose` was being called incorrectly (at least based on the docstring). Fixing this made the tests pass. I also fixed one existing test that seemed to be asserting incorrectly.

Hope that makes sense! If I'm missing something here, just let me know.

### DIFF
--- a/src/to_mdast.rs
+++ b/src/to_mdast.rs
@@ -740,7 +740,7 @@ fn on_enter_link(context: &mut CompileContext) {
 /// Handle [`Enter`][Kind::Enter]:{[`ListOrdered`][Name::ListOrdered],[`ListUnordered`][Name::ListUnordered]}.
 fn on_enter_list(context: &mut CompileContext) {
     let ordered = context.events[context.index].name == Name::ListOrdered;
-    let spread = list_loose(context.events, context.index, false);
+    let spread = list_loose(context.events, context.index, true);
 
     context.tail_push(Node::List(List {
         ordered,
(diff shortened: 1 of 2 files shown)


begleynk:
Oops accidentally closed.

Not sure what I should do about those clippy errors. Should those be addressed here?

wooorm:
*Loose* is a different concept; Spread on both the list and the items needs to be considered to know if a list is loose or not.

begleynk:
Aha I see! It didn't occur to me that the `List` itself could be spread, but that makes a lot of sense of course. For context, I'm writing an mdast -> HTML renderer.

I guess the important point is that this:
```
- foo

- bar

- baz
```

renders to
```
<ul>
<li>
<p>foo</p>
</li>
<li>
<p>bar</p>
</li>
<li>
<p>baz</p>
</li>
</ul>
```

instead of 
```
<ul>
<li>foo</li>
<li>bar</li>
<li>baz</li>
</ul>
```

Even though the `ListItem` nodes themselves are not spread.

Thanks for the explanation. I think this PR can be closed 👍

wooorm:
I recommend to look at the existing JavaScript ecosystem, which already has all you are looking for. See mdast-util-to-hast and hast-util-to-html!

Issue #108, opened 2024-02-16
Would you consider exposing a more structured error type?


begleynk:
Would you consider exposing a more structured error type that gives users direct access to e.g. the positions of the originating errors? I'm trying to provide some custom error messages to users, but the Mdx errors are all `String`s instead of something more structured.

To be more specific, I'd love to be able to give an error along these lines:

```
Expected a closing tag for `<Foo>`

1 | <Foo>
  | ^^^^^
  |   └ matching opening tag
2 |   Hello
3 | <Fo
  | ^^^
  |  └ expected closing tag
4 | More content...
```

[In this case](https://github.com/wooorm/markdown-rs/blob/60db8e5896be05d23137a6bdb806e63519171f9e/src/to_mdast.rs#L1753), the error string has the starting positions of both the opening and closing tag, but not the length/end point which would be needed to draw the lines under the offending tags.

If the error was instead something along the lines of...
```rust
pub enum Error {
  MdxClosingTagNotFound { tag: String, opening_tag_pos: Position, closing_tag_pos: Position }
  // Other error variants...
}
```
...it would allow for more customization of the errors.

The library could still have `Display` implementations for the error type for convenience and produce errors similar to how it does today.

Would you be open/interested in enabling something like this? Happy to take a look at implementing this if you'd like. I might need some pointers on where to start.

ChristianMurphy:
Welcome @begleynk!
Thanks for sharing your idea!

To preface, markdown-rs and the closely related unified js ecosystem, follow the Unix philosophy:

* Write programs that do one thing and do it well.
* Write programs to work together.
* Write programs to handle text streams, because that is a universal interface.

This module focuses on parsing valid markdown and mdx prioritizing (in order): following the CommonMark and MDX standards, being extensible to support many different use cases, and being fast and efficient.

To answer your questions:

* Is there interest in more authoring experience tools to offer localized and descriptive recommendations? **Yes**! Packages are welcome to wrap markdown-rs with this functionality
* Should that functionality live in the core parser package? **No**, the weight of a second parser to guess at intent and localizing messages goes outside the scope of markdown-rs. They would be a better fit for another module, likely a language server module like https://github.com/mdx-js/mdx-analyzer

begleynk:
Thanks for the details!

Just to clarify - I agree that putting the rendering logic for the specific kind of error messages I want to show doesn't belong into `markdown-rs`.

[...]

PR #114, opened 2024-04-22
Add error message struct

wooorm:
This is like what wooorm/mdxjs-rs#43 achieves but is implemented in the markdown parser, so that all errors can be the same struct, instead of only some.

It also follows `vfile-message`, the JavaScript counterpart, so that errors are similar in JavaScript and Rust.

Closes GH-108.
Closes: wooorm/mdxjs-rs#42.
Closes: wooorm/mdxjs-rs#43.

/cc @kdy1: this stores clean positional info on errors, is this what you want?


kdy1:
Yeap, and it looks far better than my PR. Huge thanks!

PR #120, opened 2024-08-01
Update dev-dependencies

wooorm:
Related-to wooorm/mdxjs-rs#49.

/cc @kdy1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ serde = { version = "1", features = ["derive"], optional = true }
 env_logger = "0.11"
 criterion = "0.5"
 pretty_assertions = "1"
-swc_core = { version = "0.99.6", features = [
+swc_core = { version = "0.100.0", features = [
   "ecma_ast",
   "ecma_visit",
   "ecma_parser",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,13 +26,13 @@ log = ["dep:log"]
 [dependencies]
 log = { version = "0.4", optional = true }
 unicode-id = { version = "0.3", features = ["no_std"] }
-serde = { version = "1.0", features = ["derive"], optional = true }
+serde = { version = "1", features = ["derive"], optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


wooorm:
The test that currently fails is:

https://github.com/wooorm/markdown-rs/blob/d1dc072cb64758cc973e125d8597ca57362fbb76/tests/mdx_expression_text.rs#L319-L323

kdy1:
@wooorm I merged/published https://github.com/swc-project/swc/pull/9366 but it still fails.
Are you sure that the CI failure is about `SyntaxError::UnterminatedBlockComment`?

wooorm:
Quite certain. That’s the only test case that fails.
That’s also what I reproduced as differing in 0.94 (`hi` being the eof) and 0.99 (`hi` being `1`).
But it’s about the positional info, not the name of the error.
That’s important to MDX: it needs to know when JS fails because of EOF, or because of an earlier error.

wooorm:
Made the “repro” smaller here, see last commit.
When you run:

```sh
cargo test --test mdx_expression_text -- --nocapture
```

You’ll see:

```
running 1 test
parse_file_as_expr: Err(Error { error: (1..1, Eof) })
error_end: 0..0 0 4
Error: Message { place: Some(Point(1:4 (3))), reason: "Could not parse expression with swc: Unexpected eof", rule_id: "swc", source: "mdx" }
test mdx_expression_text_gnostic ... FAILED
```

The `parse_file_as_expr:` bit prints the `1..1` part, that should be like `1..5`.
And `error_end: 0..0 0 4` should be `error_end: 0..4 4 4`

kdy1:
Sorry, I was bit busy. `Err(Error { error: (1..1, Eof) })` seems like the cause, and I think it's about `SyntaxError::Eof`
I linked the SWC repository locally, and I'll fix the span of the error soon.

kdy1:
Should be fixed by https://github.com/swc-project/swc/pull/9378

kdy1:
I updated `swc_core` to `v0.100.0`, and verified that it works

PR #121, opened 2024-08-09
Fix panic on incomplete closing frontmatter fence

nokome:
This fixes an out-of-range index panic when frontmatter only has two closing dashes e.g.

```md
---
title: Some frontmatter
--

Some other content
```

### DIFF
--- a/src/construct/document.rs
+++ b/src/construct/document.rs
@@ -565,14 +565,16 @@ fn resolve(tokenizer: &mut Tokenizer) {
                 inject_index = child_index + 1;
             }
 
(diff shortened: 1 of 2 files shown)


wooorm:
Oh, wild that this was happening! Thanks for fixing it!

wooorm:
[released](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.19), thank you!!

wooorm:
p.s. seems the `https::/stencila.io` on your GH profile is wrong? should probably be `://`?

PR #123, opened 2024-08-11
Fix padded code (text) in mdast

### LINKED ISSUE #122: mdast Inline code should strip prefix+suffix space
Per the CommonMark spec (and GH-flavored Markdown, and probably others):

> [6.1 Code spans](https://spec.commonmark.org/0.31.2/#code-spans)
> If the resulting string both begins and ends with a [space](https://spec.commonmark.org/0.31.2/#space) character, but does not consist entirely of [space](https://spec.commonmark.org/0.31.2/#space) characters, a single [space](https://spec.commonmark.org/0.31.2/#space) character is removed from the front and back

So, this markdown:

```markdown
`` hello ` ``
```

should result in `` hello ` `` (note: no space before the "hello" or after the trailing backtick).

Instead, markdown-rs (as of 1.0.0-alpha.16) renders it as ``  hello `  ``, with the spaces remaining on either side.

yshavit:
Strip one space from the beginning and end of InlineCode if (a) the value starts and ends with a space and (b) is not all spaces.

From [CommonMark spec section 6.1][1]:

> If the resulting string both begins and ends with a space character, but does not consist entirely of space characters, a single space character is removed from the front and back. 

[1]: https://spec.commonmark.org/0.31.2/#code-spans

Resolves #122

### DIFF
--- a/src/to_mdast.rs
+++ b/src/to_mdast.rs
@@ -1081,7 +1081,11 @@ fn on_exit_raw_text(context: &mut CompileContext) -> Result<(), message::Message
     }
 
     let value_bytes = value.as_bytes();
-    if value.len() > 2 && value_bytes[0] == b' ' && value_bytes[value.len() - 1] == b' ' {
+    if value.len() > 2
+        && value_bytes[0] == b' '
+        && value_bytes[value.len() - 1] == b' '
+        && !value_bytes.iter().all(|b| *b == b' ')
+    {
         value.remove(0);
         value.pop();
     }
--- a/src/to_mdast.rs
+++ b/src/to_mdast.rs
@@ -1080,6 +1080,12 @@ fn on_exit_raw_text(context: &mut CompileContext) -> Result<(), message::Message
         }
     }
 
+    let value_bytes = value.as_bytes();
+    if value.len() > 2 && value_bytes[0] == b' ' && value_bytes[value.len() - 1] == b' ' {
+        value.remove(0);
+        value.pop();
+    }
+
     match context.tail_mut() {
         Node::InlineCode(node) => node.value = value,
         Node::InlineMath(node) => node.value = value,
--- a/tests/code_text.rs
+++ b/tests/code_text.rs
@@ -205,5 +205,21 @@ fn code_text() -> Result<(), message::Message> {
         "should support code (text) as `InlineCode`s in mdast"
     );
 
+    assert_eq!(
+        to_mdast("` alpha `", &Default::default())?,
+        //     offsets: 012345678
+        Node::Root(Root {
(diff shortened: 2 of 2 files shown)


wooorm:
> whether this should apply to all inline code text, or just `` ` ``

How do you mean? Isn’t all “inline code” / “code (text)” with `` ` ``?

> whether I should adjust the position fields (and if so, which? all of them, or just the *_offsets?) to account for the stripping the spaces

I don’t think that’s needed. Positional info represents the whole, which includes the backticks too. So the spaces have no effect on it.

> apologies if this PR is not the appropriate way to do that!

This is great! Thank you!

yshavit:
> > whether this should apply to all inline code text, or just `` ` ``
> 
> How do you mean? Isn’t all “inline code” / “code (text)” with `` ` ``?

I _think_ this the raw_text also covers `$$`, but I'm not 100% — I haven't tested it out yet. And I'm also not sure whether `$$` also requires stripping the space.

> > whether I should adjust the position fields (and if so, which? all of them, or just the *_offsets?) to account for the stripping the spaces
> 
> I don’t think that’s needed. Positional info represents the whole, which includes the backticks too. So the spaces have no effect on it.

Awesome, thanks! That makes it easy.

> > apologies if this PR is not the appropriate way to do that!
> 
> This is great! Thank you!

Glad to hear it! 😁 

Unfortunately, I won't be able to get to it until later tonight (US east coast), but I'm guessing there's no urgency on this from your end. :-)

wooorm:
`$`, as in math, has no standard. I intentionally here use the exact same working of `` ` `` for `$`. So :+1: to same workings for both.

No urgency from me! I really appreciate you helping!

yshavit:
Ah, if they intentionally work the same, then perhaps this PR is ready to go? I'll go ahead and mark it as "ready for review" and subject it to full critique! :-)

src/to_mdast.rs
```diff
@@ -1080,6 +1080,12 @@ fn on_exit_raw_text(context: &mut CompileContext) -> Result<(), message::Message
         }
     }
 
+    let value_bytes = value.as_bytes();
+    if value.len() > 2 && value_bytes[0] == b' ' && value_bytes[value.len() - 1] == b' ' {
```
yshavit, line 1084: oh, actually -- I need to ensure that the value isn't _all_ space chars. If it is, they should not be stripped.

wooorm:
Thanks Yuval! [Released](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.20)!!

yshavit:
Awesome, thanks so much!

PR #127, opened 2024-08-27
Serialize mdast to markdown

### LINKED ISSUE #64: Serializing mdast to markdown
(Whoops accidentally hit enter before drafting a content for this question, my apologies for the noise!)

Hi! I have a perhaps newbie question! I can use `markdown::to_mdast` to go from `&str` -> `Node`. Is it possible/is there a function to go back to a string  - Node -> &str` - in a way that roundtrips?

I came across `Node::to_string`, and it does seem to convert nodes into a string but it also deletes the links/titles/and most other ast nodes, which if re-parsed again, results in a different ast. Unsure if this question is reasonable/within the context of this crate, but is there an alternate function elsewhere that is round-trippable to/from `&str <-> Node`? I am also happy to take a stab at implementing this "rountrippable" unparser function myself, but was wondering if a function like it already existed.

For further clarification, by "roundtripping", I would be writing a property based test, like `markdown::to_mdast(to_string(node)) == node` be true for all `node`'s.

 Thanks!

bnchi:
fixes #64 

TODO handlers : 

- [x] paragraph
- [x] text
- [x] strong
- [x] emphasis
- [x] heading
- [x] html
- [x] break
- [x] thematic-break
- [x] code
- [x] blockquote
- - [x] Complete the tests for blockquote
- [x] image
- [x] inline-code
- [x] link
- [x] list
- [x] list-item
- - [x] Add the tests for list item
- [x] root
- [x] definition
- [x] image-reference
- [x] link-reference


bnchi:
While working on this, I noticed that the JS implementation relies heavily on regular expressions. I believe we can rewrite some parts of the code in Rust instead of depending on the regex crate. Although the regex crate guarantees linear time complexity, we can eliminate the need for the regex engine in places where simpler Rust code would work. Maybe we can benchmark first but this won't happen until I move over most of the tests into Rust.

what do you think @wooorm ?

wooorm:
While the core of `markdown-rs` is `no-std + alloc`, that’s for the use case where you only really do markdown -> html, straight.
Here we are working with ASTs. I think it’s fine to depend on regexes there.
And, both JS and Rust will need to be maintained and kept in sync. So, fine to improve things if it’s fast, just, that has to be backported.
I’d recommend going with regexes first. To focus on porting the code. Instead of changing the code. That becomes confusing. Hard to manage.

As for workspace vs module, I have to particular preference currently. You?

bnchi:
Since we're working with an AST and this is a utility a workspace sounds better to me this will help to isolate the tests too. 

It's not that hard to change that decision later on, I updated the PR to use a workspace instead and kept the _no_std_ constrain.

bnchi:
I'm working on the _definition_ handler and the handler depends on https://github.com/micromark/micromark/tree/main/packages/micromark-util-decode-string to decode a markdown string, I've looked up the code in this crate and found out you've indeed ported some parts of micromark utills. 

I was wondering if it's ok to make the util from within the markdown crate pub in order to leverage functions such as `decode_named` and `decode_numeric` and add some needed constants within `util/constant ` that the `decode-string` relies on. Since this is being developed in a workspace `markdown::util` needs to be pub.

@wooorm

wooorm:
Some of the utils are already pub because they‘re used in https://github.com/wooorm/mdxjs-rs. So, having some utils pub is fine. Also, putting them in separate crates is also fine.

bnchi:
@wooorm This is ready for review more things _might_ need to be done : 

- Support tight definitions
- Support for extensions?
- Clean up the code more
- Add more tests

mdast_util_to_markdown/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "mdast_util_to_markdown"
+version = "0.0.0"
+edition = "2018"
+license = "MIT"
+
+[dependencies]
+markdown = { path = "../" }
+regex = { version = "1.7.3" }
```
wooorm, line 9:
```suggestion
regex = { version = "1" }
```

mdast_util_to_markdown/src/message.rs
```diff
@@ -0,0 +1,24 @@
...
+    }
+}
+
+impl From<&str> for Message {
+    fn from(value: &str) -> Self {
+        Message {
+            reason: value.to_string(),
+        }
+    }
+}
```
wooorm, line 24: I think there’s an error struct now in this project, since https://github.com/wooorm/markdown-rs/pull/114. Which models the JS version `vfile-message`. Could that be used here too?

wooorm (commented):
This is such an effort. Thank you! 🙇‍♂️

I didn’t check line by line, comparing JS to Rust. I’m assuming that you‘ve probably gotten that right, and trust the tests.

Some small nits.

Otherwise, looking forward to the rest of the work.
And, is there anything in particular you want me to look at?

bnchi:
> And, is there anything in particular you want me to look at?

1. if you look at the handler abstraction https://github.com/wooorm/markdown-rs/pull/127/files#diff-91a1db387b2a6664c89d595496a72eab448dfcfe854396bdef68aaaf7c0c0bdaR25 there's no `Tracker` it doesn't seem like the tracker is important in the serialization process a lot so I opt out of porting it over, are you ok with this decision ? 
2. I didn't test the performance of this implementation do you have a particular markdown file to test the JS implementation which I can use here ? 
3. https://github.com/wooorm/markdown-rs/pull/127/files#diff-f9c02e00091566ab3f89e3fd7e9f44fee067ac1ecee8afc5ee18c129431a2692R11 what do you think of the error messages that the users see ? 
4. There are a few nodes that are not serialized like tables which I think are _GFM_ and a few others, should we add them to this PR too ?
5. For handlers and extended joins we would probably need to use some sort of _dynamic dispatcher_ and I haven't thought about it much tbh, do you think they should be in this PR or should I propose a solution in an issue and make a separate PR ?
6. I've depended on Regex less in trivial places in this code if you like I can probably make a separate PR to the JS version with the same changes.
7. https://github.com/wooorm/markdown-rs/pull/127/files#diff-002c11b91d07900f0eff0a7ebe4d82aec67136f19b764c185362ed5e1e46d377 I've added an `Association` abstraction within this project, which is the same one defined in _mdast_ here https://github.com/syntax-tree/mdast?tab=readme-ov-file#association this is mainly to make the code a bit easier to read though I'm not sure if that's the right place for the abstraction or if it should be part of the _markdown_ crate.

bnchi:
By looking at the CI it doesn't seem like the tests are running for this project since it's in a separate workspace we probably need to update the CI jobs to include _mdast_util_to_markdown_ in the pipeline.

mdast_util_to_markdown/src/message.rs
bnchi: Yes sure

mdast_util_to_markdown/src/state.rs
```diff
@@ -27,9 +27,9 @@ use regex::{Captures, Regex, RegexBuilder};
 
 #[derive(Debug)]
 enum Join {
```
bnchi, line 29: I've created these join variants name out of my understating for how _join_ works, in JS the returned type are a bit confusing tbh and I would rather have these to describe the type of a _Join_  in Rust

mdast_util_to_markdown/src/message.rs
bnchi: I updated it let me know if the changes are fine here

bnchi:
I updated the CI jobs and added some missing tests I've also did some refactoring, I don't think I'm going to be adding anything new to this PR unless there's some required changes. I might make a follow up PR to improve the performance if possible or if I spotted some bugs.

For un-handled nodes which are GFM nodes like tables and the others not sure if I should add the handlers here or in some other project because the JS implementation doesn't seem to have them.

Thanks !

wooorm:
Thanks for your contributed work!
I remember that there were some questions you still had, about the tracker for example, but can’t seem to find it now. I was still supposed to answer that.
As for the tracker: that isn’t used much in the defaults currently. But it is needed to serialize JSX. In the future, with this tracker, `mdast-util-to-markdown` itself could become smarter and wrap lines at 80 characters for example. That now only works for JSX (from MDX).

As for the Node extensions, and whether they should be supported here: I do think they should be here.
In JavaScript, code size is very important. And plugability of extensions into some main thing is easy.
In rust, code size is less important, and plugability is very hard.
That is why GFM and frontmatter and MDX as syntax extensions are also in `markdown-rs`, instead of external packages.

mdast_util_to_markdown/src/configure.rs
```diff
@@ -0,0 +1,50 @@
...
+            rule: '*',
+            strong: '*',
+            close_atx: false,
+            rule_spaces: false,
+            resource_link: false,
+            setext: false,
+            tight_definitions: false,
+        }
+    }
+}
```
wooorm, line 50: Needs some docs. Could be taken from `mdast-util-to-markdown` as well. See core of `markdown-rs` for info on how to do docs.

mdast_util_to_markdown/src/construct_name.rs
```diff
@@ -0,0 +1,30 @@
...
+    LinkReference,
+    List,
+    ListItem,
+    Paragraph,
+    Phrasing,
+    Reference,
+    Strong,
+    TitleApostrophe,
+    TitleQuote,
+}
```
wooorm, line 30: Same, probably all `pub` things could use some docs

wooorm (commented):
It looks pretty good.

Can you confirm whether you took over all the tests, without change to input/output?

I’ll look at the `src` folder locally after merge to compare JS to RS just to be sure: from a quick glance, it looks good :+1:

bnchi:
> Can you confirm whether you took over all the tests, without change to input/output?

I did move over all the tests for the handlers except for core and round-trips tests. 

I made sure to move the tests by hand and didn't trust an LLM to do the conversion as I initially planned. Since I did it by hand I can guarantee that 90% of the tests are an exact match though I might have missed something. I can spend some times to do an audit after a merge or if there's anyone from the community that can help with reviewing the tests that will be very much appreciated!

One handler tests which I didn't have the time to complete is list item I worked on that today and I just moved what seem to be a hit for every code point here : https://github.com/wooorm/markdown-rs/pull/127/commits/9bfa391ca9027913278685ee89af2741e10dccb8#diff-5b1c4c5fdf4be7bbe31465522517f284b165d6370d4c5ff954df594855a9869a

mdast_util_to_markdown/src/configure.rs
bnchi: I've added comments here let me know if that's good enough

bnchi:
I completed copying over the tests so now all of the tests are a match, I also did a quick review everything looks fine to me, defiantly a few performance improvements and refactors can be done in a separate PR but now since the tests are moved over it's easy to make changes without worrying about something breaking.

I've also added comments as proposed for some data structures.

bnchi:
wooorm knows more about this than me, but I think Yaml and Toml are not part of the _CommonMark_ specs they're part of the _frontmatter_ extension this PR ports over the JS implementation here [mdast-util-to-markdown ](https://github.com/syntax-tree/mdast-util-to-markdown)

In order for Yaml and Toml to work we need to add support for frontmatter serialization I believe the JS version is here [mdast-util-frontmatter](https://github.com/syntax-tree/mdast-util-frontmatter)

ChristianMurphy:
Some extensions have built in parsers that can be turned on with an option.
Including frontmatter! https://github.com/wooorm/markdown-rs?tab=readme-ov-file#extensions

It could make sense to support serialization for those tools.
But this PR is pretty sizable already.
I'd lean towards having extensions in follow up PRs.

ChristianMurphy:
Also additional testing, using a fuzzer to to validate that content can round trip.

In other words markdown can be parsed to a tree, be serialized again, and parsed again should have the same tree structure (though positions and exact characters may change).

That technique found a lot of quirks in the original micromark/remark parser and serialized.
And could be valuable to repeat here.
But is likely outside the scope of this PR.

References:
About fuzzing: https://rust-fuzz.github.io/book/
Current parser fuzz test: https://github.com/wooorm/markdown-rs/tree/main/fuzz

wooorm:
> But this PR is pretty sizable already.
> I'd lean towards having extensions in follow up PRs.

Right yes, that indeed. Somewhere above @Bnchi and I talked about this too and concluded the same.

I believe @Bnchi has used up a lot of the time they have to work on this. So if @augustkline or others are up for follow up PRs, I’d love that!

wooorm (approved):

wooorm:
If I run `cargo clippy --all-features --all-targets --workspace`, I get:

```
error[E0599]: no method named `len` found for struct `regex::Match` in the current scope
  --> mdast_util_to_markdown/src/handle/inline_code.rs:53:47
   |
53 |                         && &value[position..m.len()] == "\n"
   |                                               ^^^ method not found in `Match<'_>`

error[E0599]: no method named `len` found for struct `regex::Match` in the current scope
   --> mdast_util_to_markdown/src/state.rs:131:62
    |
131 |                         .map(|captured_group| captured_group.len())
    |                                                              ^^^ method not found in `Match<'_>`

error[E0599]: no method named `len` found for struct `regex::Match` in the current scope
   --> mdast_util_to_markdown/src/state.rs:498:53
    |
498 |             start = full_match.start() + full_match.len();
    |                                                     ^^^ method not found in `Match<'_>`
```

What am I missing 🤔

wooorm:
Only thing perhaps is some docs in this readme here?
And, how should this new code be used? Should it be published? As alpha?

Enoumy:
Thank you so much for writing this PR!

bnchi:
> Only thing perhaps is some docs in this readme here?

Yeah I can add instructions on how to use this code in a readme

> Should it be published? As alpha?

Alpha sounds alright, what do you think ?

wooorm:
Sounds good to me! :+1:

766F6964:
@bnchi I am currently playing around with this, since I need a pipeline:
_markdown data _ -> _markdown ast_ -> _transformations on markdown ast_ -> back to markdown

I got this setup working, however, something that appears to be completely missing are tables?

bnchi:
Tables are part of the GFM extension in order to add it we will need to
support :  https://github.com/syntax-tree/mdast-util-gfm

766F6964:
@bnchi Is this planned? Or is it already being worked on?

wooorm:
neither: you can work on this

PR #135, opened 2024-09-15
Fix serde, change `BlockQuote` -> `Blockquote`

### LINKED ISSUE #72: MDAST: serde Serialization/Deserialization not working

h7kanna:
* Fixed #72 
* Added a sanity test for the same

**NOTE** : The included sanity test makes sure that at-least the AST serializes and deserializes. More rigorous tests should be added later. Currently, it works as the construct names 'camelCased' match the [Mdast JSON](https://github.com/syntax-tree/mdast#nodes) except for Blockquote (changing it will be API change). 
As it is still in alpha changing 'BlockQuote' enum to 'Blockquote' will make serde work completely.

**NOTE**: This PR makes another pending one obsolete https://github.com/wooorm/markdown-rs/pull/74

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -254,18 +254,15 @@ The following bash scripts are useful when working on this project:
     ```
 *   format:
     ```sh
-    cargo fmt && cargo fix --all-targets
+    cargo fmt && cargo fix --all-targets --all-features
     ```
 *   lint:
     ```sh
-    cargo fmt --check && cargo clippy --examples --tests --benches --all-features
+    cargo fmt --check && cargo clippy --examples --tests --benches --all-features --all-features
     ```
 *   test:
     ```sh
-    RUST_BACKTRACE=1 cargo test
-    ```
-    ```sh
-    RUST_BACKTRACE=1 cargo test --features json
+    RUST_BACKTRACE=1 cargo test --all-features
     ```
 *   docs:
     ```sh
--- a/readme.md
+++ b/readme.md
@@ -264,6 +264,9 @@ The following bash scripts are useful when working on this project:
     ```sh
     RUST_BACKTRACE=1 cargo test
     ```
+    ```sh
+    RUST_BACKTRACE=1 cargo test --features json
+    ```
 *   docs:
     ```sh
     cargo doc --document-private-items
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,5 +20,5 @@ jobs:
       - uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
-      - run: cargo install cargo-tarpaulin && cargo tarpaulin --out xml
+      - run: cargo install cargo-tarpaulin && cargo tarpaulin --features json --out xml
       - uses: codecov/codecov-action@v4
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ serde = { version = "1", features = ["derive"], optional = true }
 env_logger = "0.11"
 criterion = "0.5"
 pretty_assertions = "1"
+serde_json = { version = "1" }
 swc_core = { version = "0.100", features = [
   "ecma_ast",
   "ecma_visit",
--- a/src/mdast.rs
+++ b/src/mdast.rs
@@ -86,7 +86,7 @@ pub enum AlignKind {
 #[cfg_attr(
     feature = "serde",
     derive(serde::Serialize, serde::Deserialize),
-    serde(tag = "type", rename = "type")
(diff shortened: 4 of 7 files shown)


readme.md
```diff
@@ -264,6 +264,9 @@ The following bash scripts are useful when working on this project:
     ```sh
     RUST_BACKTRACE=1 cargo test
     ```
+    ```sh
+    RUST_BACKTRACE=1 cargo test --features json
+    ```
```
wooorm, line 269: Do you recommend that humans run both commands? One or the other?

wooorm:
> (changing it will be API change).
As it is still in alpha changing 'BlockQuote' enum to 'Blockquote' will make serde work completely.

Are you saying that currently (main branch) the `type` field is *not* `blockquote` (the correct value according to mdast)?
And that this branch doesn’t change that either?
I would prefer to fix it: to use `blockquote`.

---

Are you interested in adding, as you note in your OP, some more tests?

readme.md
h7kanna: Can be replaced with 'cargo test --all-features'

wooorm: ok, can you do that? One test example, with `all-features`?

wooorm:
> invalid json 

I don’t think JSON has a limitation on non-duplicate fields? https://www.json.org/json-en.html.

> individually merged

fine!


***


Changing Block Quote is :+1: for me!

readme.md
h7kanna: done

h7kanna:
> > invalid json

I mean deserialization failing because of above duplicate 'type'.
> 
> I don’t think JSON has a limitation on non-duplicate fields? https://www.json.org/json-en.html.
> 
> > individually merged
> 
> fine!
> 
> Changing Block Quote is 👍 for me!

Yes, made the change as well.

wooorm:
> I mean deserialization failing because of above duplicate 'type'.

I know, but I am trying to point out that I do not think this is not something JSON defines. I think it is something that some libraries do: in JS, `JSON.parse('{"a":1,"a":2}')` is `{a:2}`.

wooorm (commented):
Looks good. Are you planning to add some more tests in this PR, or separately?

h7kanna:
> Looks good. Are you planning to add some more tests in this PR, or separately?

Definitely adding more test cases. I will do in follow up PR. It is just taking time to write down the fixtures. Most of the content is just spans and points.

wooorm:
Thanks!

wooorm:
Looking forward to your other PR!

wooorm:
@h7kanna Do you perhaps have an idea how to fix this?

https://github.com/wooorm/markdown-rs/blob/4ad5342156cf5f5256fc452c4583be99a3e80412/src/mdast.rs#L81-L82

It should serialize in JSON as `null`. Not `"none"`.

PR #137, opened 2024-09-17
Added tests for Serde

h7kanna:
This is follow up for PR https://github.com/wooorm/markdown-rs/pull/135
Without [jaq](https://github.com/01mf02/jaq) tests are becoming verbose and unwieldy. 
 I hope adding a test dependency is fine with you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,9 @@ serde = { version = "1", features = ["derive"], optional = true }
 [dev-dependencies]
 env_logger = "0.11"
 criterion = "0.5"
+jaq-core =  { version = "2.0.0-alpha" }
+jaq-json = { version = "1.0.0-alpha", features = ["serde_json"] }
+jaq-std = { version = "2.0.0-alpha" }
 pretty_assertions = "1"
 serde_json = { version = "1" }
 swc_core = { version = "0.100", features = [
--- a/tests/serde.rs
+++ b/tests/serde.rs
@@ -1,17 +1,16 @@
-use markdown::mdast::Node;
-use markdown::message::Message;
+mod test_utils;
 
 #[allow(unused)]
 #[derive(Debug)]
 enum Error {
-    Mdast(Message),
+    Assert(String),
+    Mdast(markdown::message::Message),
     Serde(serde_json::Error),
 }
 
 #[cfg_attr(feature = "serde", test)]
-fn serde() -> Result<(), Error> {
-    let source = markdown::to_mdast(
-        r#"---
+fn serde_root() -> Result<(), Error> {
+    let source = r#"---
 title: Serde
 ---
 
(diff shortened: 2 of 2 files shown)


wooorm:
Hmm. I am not sure whether I want tests that only look for a `type` field?
What was verbose and unwieldy?
Testing the generated JSON doesn’t have to be too complex?

h7kanna:
> Thanks! Hmm. I am not sure whether I want tests that only look for a `type` field? What was verbose and unwieldy? Testing the generated JSON doesn’t have to be too complex?

Not complex, very verbose and unreadable tests. I can add asserts for all the properties not only for the type if you are ok with the approach. But by directly testing JSON, most of the content of the fixture is spans (which I think redundant to check for serde).

For example 
```
assert_jq("a", ".children[0].type", "paragraph", None)
assert_jq("a", ".children[0].value", "a", None)
```
vs

```
assert_eq!(r#"{
  "type": "root",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "text",
          "value": "a",
          "position": {
            "start": {
              "line": 1,
              "column": 1,
              "offset": 0
            },
            "end": {
              "line": 1,
              "column": 2,
              "offset": 1
            }
          }
        }
      ],
      "position": {
        "start": {
          "line": 1,
          "column": 1,
          "offset": 0
        },
        "end": {
          "line": 1,
          "column": 2,
          "offset": 1
        }
      }
    }
  ],
  "position": {
    "start": {
      "line": 1,
      "column": 1,
      "offset": 0
    },
    "end": {
      "line": 1,
      "column": 2,
      "offset": 1
    }
  }
}"#, actual_serde_json)

```

wooorm:
If you remove positional info and indentation it becomes fine?

h7kanna:
> If you remove positional info and indentation it becomes fine?

Do you mean like below? 

```
assert_eq!(r#"{
  "type": "root",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "text",
          "value": "a",
        }
      ]
    }
  ]
}"#, actual_serde_json)
```

wooorm:
What I was thinking of is:

```
assert_eq!(r#"{"type": "root","children": [{"type": "paragraph","children": [{"type": "text","value": "a"}]}]}"#, actual_serde_json)
```

Or, isn’t it `actual` first, then `expected`?:

```
assert_eq!(actual_serde_json, r#"{"type":"root","children":[{"type": "paragraph","children":[{"type":"text","value":"a"}]}]}"#)
```

You can also choose to dig into a node of course: we often don’t care about `root`:

```
assert_eq!(node, r#"{"type":"paragraph","children":[{"type":"text","value":"a"}]}"#)
```

h7kanna:
> What I was thinking of is:
> 
> ```
> assert_eq!(r#"{"type":"root","children":[{"type":"paragraph","children":[{"type":"text","value":"a"}]}]}"#, actual_serde_json)
> ```
> 
> Or, isn’t it `actual` first, then `expected`?:
> 
> ```
> assert_eq!(actual_serde_json, r#"{"type":"root","children":[{"type":"paragraph","children":[{"type":"text","value":"a"}]}]}"#)
> ```
> 
> You can also choose to dig into a node of course: we often don’t care about `root`:
> 
> ```
> assert_eq!(node, r#"{"type":"paragraph","children":[{"type":"text","value":"a"}]}"#)
> ```

I mean we have to come up the the implementation of that `assert_eq` which I did using jql.
Direct JSON(expected) comparison without the positional info is not possible as parsed MDAST(actual) will always have this

wooorm:
> Direct JSON(expected) comparison without the positional info is not possible as parsed MDAST(actual) will have the positions.

I believe that the comparison is possible after using a utility function?

This is the JavaScript version: https://github.com/syntax-tree/unist-util-remove-position. Walking trees is a bit different in Rust, but the approach should be similar?

h7kanna:
> > Direct JSON(expected) comparison without the positional info is not possible as parsed MDAST(actual) will have the positions.
> 
> I believe that the comparison is possible after using a utility function? Yes
> 
> This is the JavaScript version: https://github.com/syntax-tree/unist-util-remove-position. Walking trees is a bit different in Rust, but the approach should be similar?
I will try to implement visit on Node.

h7kanna:
Closing this one as there two many changes.
Follow up with requested changes here https://github.com/wooorm/markdown-rs/pull/140

PR #138, opened 2024-09-19
Fix parsing of JSX with expression

### LINKED ISSUE #136: Unable to parse inline JSX element with MDX expression child
It seems that since `1.0.0-alpha.17` this syntax doesn't work any more:

```mdx
<span>{1}</span>
```

You get the following error:
```rust
Err(
    Message {
        place: Some(
            Point(
                2:1 (17),
            ),
        ),
        reason: "Expected a closing tag for `<span>` (1:1)",
        rule_id: "end-tag-mismatch",
        source: "markdown-rs",
    },
)
```

`1.0.0-alpha.16` parses it correctly:
```rust
Ok(
    Root {
        children: [
            Paragraph {
                children: [
                    MdxJsxTextElement {
                        children: [
                            MdxTextExpression {
                                value: "1",
                                position: Some(
                                    1:7-1:10 (6-9),
                                ),
                                stops: [
                                    (
                                        0,
                                        7,
                                    ),
                                ],
                            },
                        ],
                        position: Some(
                            1:1-1:17 (0-16),
                        ),
                        name: Some(
                            "span",
                        ),
                        attributes: [],
                    },
                ],
                position: Some(
                    1:1-1:17 (0-16),
                ),
            },
        ],
        position: Some(
            1:1-2:1 (0-17),
        ),
    },
)
```

I used the following config:
```rust
    let mut opts = markdown::Options::default();
    let mut parse_opts = markdown::ParseOptions::mdx();
    parse_opts.constructs.gfm_table = true;

    opts.parse = parse_opts;

    // Parse the markdown file into an AST
    let mdast = markdown::to_mdast(&content, &opts.parse);
```

Perhaps this is the commit that caused it? https://github.com/wooorm/markdown-rs/commit/71361e4ada574d50cd9880422f5cf7ca3a38689c

Happy to take a stab at fixing this. Any pointers about what might be going on and where to start would be appreciated!

begleynk:
**Disclaimer:** I am not entirely sure that this is the correct fix, but it does pass the tests. Happy to make changes in the likely case I haven't quite understood this fully.

This fixes the issues mentioned in #136, where this case `<x>{1}</x>` was no longer parsed, and `{1}<x/>` caused a panic when converted into an AST (not an issue when converting to HTML).

I was first confused about what the expected output should be here, so I looked at micromark for inspiration. Here's what I found:

- Micromark parses `<x>{1}</x>` as a block level Flow element with an expression (unlike this crate as of alpha.16, which produced a paragraph with an inline JSX text node)
- Micromark parses `{1}</x>` into an AST with a `root` element, with two children: the expr, and the JSX tag

This change now makes this crate matches micromark's behavior.

<details><summary>Test program I used</summary>

```js
import {mdxjs} from 'micromark-extension-mdxjs'
import {fromMarkdown} from 'mdast-util-from-markdown'
import {mdxFromMarkdown} from 'mdast-util-mdx'


const doc = "<span>{1}</span>";

const tree = fromMarkdown(doc, {
  extensions: [mdxjs()],
  mdastExtensions: [mdxFromMarkdown()]
})


console.log(JSON.stringify(tree, null, 2))
```
</details>

This also removes the `MdxExpressionFlowNok` part of the state, as it wasn't used any more.

Fixes #136

### DIFF
--- a/src/construct/mdx_expression_flow.rs
+++ b/src/construct/mdx_expression_flow.rs
@@ -150,9 +150,10 @@ pub fn end(tokenizer: &mut Tokenizer) -> State {
         Some(b'<') if tokenizer.parse_state.options.constructs.mdx_jsx_flow => {
             // We can’t just say: fine.
             // Lines of blocks have to be parsed until an eol/eof.
+            tokenizer.tokenize_state.token_1 = Name::MdxJsxFlowTag;
             tokenizer.attempt(
-                State::Next(StateName::MdxExpressionFlowAfter),
-                State::Next(StateName::MdxExpressionFlowNok),
+                State::Next(StateName::MdxJsxFlowAfter),
+                State::Next(StateName::MdxJsxFlowNok),
             );
             State::Retry(StateName::MdxJsxStart)
         }
@@ -169,17 +170,6 @@ pub fn end(tokenizer: &mut Tokenizer) -> State {
             State::Nok
         }
     }
-}
-
-/// At something that wasn’t an MDX expression (flow).
-///
-/// ```markdown
-/// > | {A} x
-///     ^
-/// ```
-pub fn nok(tokenizer: &mut Tokenizer) -> State {
-    reset(tokenizer);
-    State::Nok
 }
 
 /// Reset state.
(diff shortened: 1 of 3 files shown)


wooorm (approved):

wooorm:
thanks!

PR #148, opened 2024-10-09
Add support for serializing math nodes as markdown

bnchi:
Adds support for serializing math nodes in mdast_util_to_markdown

TODO : 

- [x] Add support for inline math and complete the tests


bnchi:
@wooorm I made this a draft to get your opinion on how we're going to be serializing the extensions should we do any special way in the API by allowing the user to configure the extension or no. This draft doesn't allow the user to configure the extension and we serialize the extensions "nodes" in a normal way.

wooorm:
> should we do any special way in the API by allowing the user to configure the extension or no

No, I don’t think so. I think it’s good to always serialize nodes, if we understand them!

mdast_util_to_markdown/src/configure.rs
```diff
@@ -92,6 +96,7 @@ impl Default for Options {
             setext: false,
             strong: '*',
             tight_definitions: false,
+            single_dollar_text_math: true,
         }
     }
 }
```
wooorm, line 102: Can you sort both objects?

mdast_util_to_markdown/src/construct_name.rs
```diff
@@ -230,4 +230,24 @@ pub enum ConstructName {
...
+    MathFlow,
+    /// Math (flow) meta flag.
+    ///
+    /// ```markdown
+    /// > | $$a
+    ///       ^
+    ///   | b
+    ///   | $$
+    /// ```
+    MathFlowMeta,
```
wooorm, line 252: Same, can be sorted

mdast_util_to_markdown/src/handle/root.rs
```diff
@@ -35,6 +35,7 @@ fn phrasing(child: &Node) -> bool {
             | Node::Emphasis(_)
             | Node::ImageReference(_)
             | Node::Image(_)
+            | Node::InlineMath(_)
```
wooorm, line 38: Same, sort :)

mdast_util_to_markdown/src/configure.rs
```diff
@@ -65,6 +65,10 @@ pub struct Options {
     /// Setext headings cannot be used for empty headings or headings with a
     /// rank of three or more.
     pub setext: bool,
+    /// (`bool`, default: `true`) — whether to support math (text) with a single dollar.
```
wooorm, line 68:
```suggestion
    /// Whether to support math (text) with a single dollar (`bool`, default: `true`).
```

I place this at the end of the sentence

mdast_util_to_markdown/src/configure.rs
```diff
@@ -65,6 +65,10 @@ pub struct Options {
...
+    /// Single dollars work in Pandoc and many other places, but often interfere with “normal”
+    /// dollars in text. If you turn this off, you can still use two or more dollars for text math
```
wooorm, line 70:
```suggestion
    /// dollars in text.
    /// If you turn this off, you can still use two or more dollars for text math
```

And then I try to break after sentences.
the reasoning for that is roughly https://sembr.org: it improves diffing and reasoning.

mdast_util_to_markdown/src/handle/inline_math.rs
```diff
@@ -0,0 +1,82 @@
+//! JS equivalent: https://github.com/syntax-tree/mdast-util-math/blob/main/lib/index.js#L241
```
wooorm, line 1: :+1:

wooorm (commented):
Looks good, just a small nit on docs :)

Otherwise, I think it’s good? You? Anything in particular you wanted me to look at?

mdast_util_to_markdown/src/configure.rs
bnchi: Oh IC, I usually use a line in my editor to indicate when to break, this is interesting though thanks for letting me know.

bnchi:
> Otherwise, I think it’s good? You? Anything in particular you wanted me to look at?

No, all good. 

I'll probably work on the mdx nodes next. Thank you :)

wooorm:
Super, thank you! 👏

SamPeng87:
so,I think that to_markdown api has need some time to used?thanks that pr，I need that feature too

bnchi:
@SamPeng87 It's usable but it needs some documentations and most likely a release into the registry, for now you can add the following to your `Cargo.toml` and use it : 

```toml
[dependencies]
mdast_util_to_markdown = { git = "https://github.com/wooorm/markdown-rs.git" }
```

The api is pretty simple just have a look at the exposed function here : https://github.com/wooorm/markdown-rs/blob/main/mdast_util_to_markdown/src/lib.rs

SamPeng87:
@bnchi yes,thanks，I found that there was no document to tell me how to use it, but I found that there was in master branch, so I just studied the latest master of github master branch instead of crate.io。that work。thanks you。

PR #153, opened 2024-10-20
Add mdx serialization support

bnchi:
- [ ] Add support for mdx_jsx and move over the tests

This PR is dependent on #152


bnchi:
@wooorm 

This is roughly 60% finished, but I want to take your opinion on the way of how you're doing tracking and some of the (monkey patches) that are in the js version for `mdx-js`.

In _mdx-js_ there's a copy of the `containerFlow` function here https://github.com/syntax-tree/mdast-util-mdx-jsx/blob/main/lib/index.js#L709

It's a modified copy which seems to have more logic in it for indent, but it's only called from within the _mdx-jsx_ module, my question here is does that mean _mdx-jsx_ doesn't care about any of the previous tracking that have happened while we were serializing the tree ? If it's not the case do you think we should add support for tracking in _mdast-util-to-markdown_?

If we did so the Rust version might be out of sync with the JS one, but we can probably also make the same PR in the JS version, and I honestly don't know how many of the tests are going to change (if any).

There are some other overridden options when we have a tree with _jsx_ nodes here https://github.com/syntax-tree/mdast-util-mdx-jsx/blob/main/lib/index.js#L537

Does that mean in the Rust version at run time we need to know if we have a _jsx_ node before we start serializing ? Or should we let the user configure this option by themselves and error or panic when we find a _jsx_ node with the wrong configuration. 

We can be explicit about the second option in the docs, what do you think ?

bnchi:
I don't have the bandwidth to address any of the issues in this PR if someone would like to build on top of this draft feel free

PR #154, opened 2024-10-20
Fix clippy issues

bnchi:
With newer versions of Rust and Clippy the CI jobs errors out, this PR fixes the errors raised by Clippy to make the CI pass.


bnchi:
There are conflicts in this PR which I think means the failing jobs were addressed by the maintainer 

cc @wooorm

wooorm:
Right, yup, thanks!

Issue #163, opened 2025-02-20
Data URIs


lovasoa:
Hello !

  I am the maintainer of [SQLPage](https://sql-page.com), which embeds markdown-rs.
  Recently, an user reported that image data-URIs do not work in markdown: https://github.com/sqlpage/SQLPage/discussions/803#discussioncomment-12260619

  Is this on purpose ? Could image data-URI support be added ?

wooorm:
hi! yes, they are unsafe. see https://github.com/wooorm/markdown-rs#security!

lovasoa:
Oh, sorry for opening the issue without checking that first.

Any reason why you consider data URLs unsafe ?

lovasoa:
I understand why they could be considered unsafe in the context of a link, but what about images ?

wooorm:
Images don’t need interaction, those `src`s run. A quick duckduckgo gives me stuff like https://wiki.zacheller.dev/web-app-pentest/xss/data-urls

lovasoa:
> So, if we want to run JS we can actually inject either of these into a URL:
> ```
> data:text/html,<script>alert('hi');</script>
> data:text/javascript,alert(1)  
> ```
>  - https://wiki.zacheller.dev/web-app-pentest/xss/data-urls

This is wrong and easily debunked. Just paste the following to an html file and open it : 

```html
<img src="data:text/javascript,alert(1)" />
```

The javascript is not executed.

If we allow http URLs pointing to any domain, then allowing data URIs does not increase the attack surface in any way. 
Anything included in a data URI could just as well be returned by a remote attacker-controlled server in a normal HTTP URL.

wooorm:
It’s also something that GH + the sanitizers I know do, so I’m doubtful that it suddenly is completely safe

lovasoa:
The reason github does it is probably related to caching. 

On github, the following markdown:

```markdown
![](http://example.com)
```

is rendered as

```html
<img src="https://camo.githubusercontent.com/7f71acf223968ffae7687e6452f172b5744538e32b3752ca11744db019aa1c16/687474703a2f2f6578616d706c652e636f6d" alt="" data-canonical-src="http://example.com" style="max-width: 100%;">
```

Github caches images on their own servers, and do not want to overload pages with large data URIs.

But in a standalone markdown rendering library, there is no reason to consider image data URIs as *dangerous*.

wooorm:
[...]

PR #164, opened 2025-02-20
Allow data URIs as image sources

lovasoa:
Data URIs allow embedding image data directly within HTML documents using the `src` attribute of `` tags. This practice is generally safe and does not lead to arbitrary code execution. 

I understand this is a scary pull request to merge when one is not familiar with the topic, so here is a detailed explanation:

1. **Image Rendering Context**: When a Data URI is used in an `` tag, browsers interpret and render the content strictly as an image. They do not execute any embedded scripts or code within this context. 
  - https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image#restrictions
  - https://www.w3.org/Graphics/SVG/IG/resources/svgprimer.html#SVG_image

2. **SVG Restrictions**: SVGs can contain scripts, all browsers impose security restrictions. SVG images embedded via `` tags have scripting disabled, preventing any embedded JavaScript from executing.

Using Data URIs within `` tags is a secure method for embedding images directly into HTML documents. Browsers handle these URIs in a way that prevents code execution. 

Other markdown rendering libraries, including the one used by github (comrak), do not consider image data URLs as unsafe.

Example:

```rust
use comrak::*;

fn main() {

let mut options = Options::default();
let input = "
![](http://example.com)
![](data:image/png,aaa)
![](javascript:dangerous)
";

println!("{}", markdown_to_html(input, &options));
}
```

Results in

```html
<p><img src="http://example.com" alt="" />
<img src="data:image/png,aaa" alt="" />
<img src="" alt="" /></p>
```

fixes https://github.com/wooorm/markdown-rs/issues/163

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -285,7 +285,7 @@ The following bash scripts are useful when working on this project:
 The typical security aspect discussed for markdown is [cross-site scripting
 (XSS)][xss] attacks.
 Markdown itself is safe if it does not include embedded HTML or dangerous
-protocols in links/images (such as `javascript:` or `data:`).
+protocols in links/images (such as `javascript:`).
 `markdown-rs` makes any markdown safe by default, even if HTML is embedded or
 dangerous protocols are used, as it encodes or drops them.
 Turning on the `allow_dangerous_html` or `allow_dangerous_protocol` options for
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -518,9 +518,9 @@ pub struct CompileOptions {
     ///
     /// URLs that have no protocol (which means it’s relative to the current
     /// page, such as `./some/page.html`) and URLs that have a safe protocol
-    /// (for images: `http`, `https`; for links: `http`, `https`, `irc`,
+    /// (for images: `http`, `https`, `data`; for links: `http`, `https`, `irc`,
     /// `ircs`, `mailto`, `xmpp`), are safe.
-    /// All other URLs are dangerous and dropped.
+    /// All other URLs are considered dangerous by this library and dropped.
     ///
     /// ## Examples
     ///
--- a/src/util/constant.rs
+++ b/src/util/constant.rs
@@ -274,7 +274,7 @@ pub const SAFE_PROTOCOL_HREF: [&str; 6] = ["http", "https", "irc", "ircs", "mail
 /// List of protocols allowed, when operating safely, as `src` on `img`.
 ///
 /// This list is based on what is allowed by GitHub.
-pub const SAFE_PROTOCOL_SRC: [&str; 2] = ["http", "https"];
+pub const SAFE_PROTOCOL_SRC: [&str; 3] = ["http", "https", "data"];
 
 /// The number of characters that form a tab stop.
 ///
--- a/tests/misc_dangerous_protocol.rs
+++ b/tests/misc_dangerous_protocol.rs
@@ -113,6 +113,12 @@ fn dangerous_protocol_image() {
         "<p><img src=\"a/b:c\" alt=\"\" /></p>",
         "should allow a colon in a path"
     );
+
+    assert_eq!(
+        to_html("![](data:image/png;base64,abc)"),
+        "<p><img src=\"data:image/png;base64,abc\" alt=\"\" /></p>",
+        "should allow data URIs"
+    );
 }
 
 #[test]


lovasoa:
Hello! 
 
  SVG is indeed the only image format supported in img tags that supports scripting, so I thought it was important to clarify that svg scripts are not executed. Non-image formats are just ignored, and other supported image formats cannot include scripts. 

  About allowing all protocols on images: I don't see any legitimate use case for javascript:  or other protocols in image sources, but this is indeed not a security issue. We could add it, but I'm not sure that would be useful. 

  Advising users to completely disable the built-in protocol validation to allow the perfectly safe data URLs is not a great security posture. 

 About GitHub: as I showed, the markdown parsing and rendering library they use does allow data URLs. They indeed have a filter removing data URLs further in their pipeline, but that's still something that could be implemented with this library. 

The point is, there is absolutely no security reason to disable data URLs, or to put them behind the same flag that for instance disables JavaScript: URLs in links.

wooorm:
> SVG is […] the only image format supported in img tags […]. Non-image formats are just ignored, and other supported image formats cannot include scripts.

I believe that the HTML spec does not specify a particular list of image formats.
I was under the impression that PDFs for example were also shown.
I believe the spec also does not specify whether scripting should be allowed in them.
Or, `data:`.
The question is thus: could there be a browser that supports some image format that allows scripting (in the future, in the past)?

Your argumentation also begs the question: Is currently maintained browsers the target of this project? Only popular ones? Are browsers the only target or are there other user agents?

I can at least confirm that [Opera from 2012 had scripting in `src`](https://github.com/cure53/DOMPurify/issues/205#issuecomment-286042087) (see also [this](https://github.com/cure53/DOMPurify/issues/911#issuecomment-1952274473)).

> but this is indeed not a security issue. We could add it, but I'm not sure that would be useful.

I understand your argumentation to stretch against all allow-lists for `src`; you believe `src` to be always safe.
I do not understand why you would then push for a half solution. Would it not be simpler to push for allowing all values on `src`?

> Advising users to completely disable the built-in protocol validation to allow the perfectly safe data URLs is not a great security posture.

I am not advising that. I am advising an allow list of `http` and `https`. I think that that is enough. For more complex needs, users can use their own sanitizing algorithm. Especially, when plugins are there.

Similar argumentation to `data:` on `src` of `img` can be used for `ftp:`. Or authors writing `<i>` and `</i>`. Those are also known to be safe. But, things get complex quickly.

> as I showed

Comrak is as far as I am aware still not used by GH. GH uses `cmark-gfm`:

* https://github.com/github/cmark-gfm/blob/587a12bb54d95ac37241377e6ddc93ea0e45439b/src/html.c#L404
* https://github.com/github/cmark-gfm/blob/587a12bb54d95ac37241377e6ddc93ea0e45439b/src/scanners.re#L344

Though, I do believe they run this in dangerous mode, and then filter later with a tool *similar* but *different* to the [HTML pipeline](https://github.com/github/html-pipeline/blob/97e5e97d85e1158ab4c0db2a0b9cbff6486ae8a1/lib/html/pipeline/sanitization_filter.rb#L77-L87).

See also https://github.com/kivikakk/comrak/pull/169. It looks like comrak is also doing things differently than what you propose.

lovasoa:
>  The question is thus: could there be a browser that supports some image format that allows scripting (in the future, in the past)?
> Your argumentation also begs the question: Is currently maintained browsers the target of this project? Only popular ones? Are browsers the only target or are there other user agents?

All image formats are already allowed by this library, as long as they are transmitted through HTTP or HTTPS. Allowing data URLs does not increase the attack surface.

> I understand your argumentation to stretch against all allow-lists for src; you believe src to be always safe.
I do not understand why you would then push for a half solution. Would it not be simpler to push for allowing all values on src?

I would be perfectly fine with that ! Do you want me to remove the protocol restriction completely ?

> Comrak is as far as I am aware still not used by GH. 

I think github uses https://github.com/gjtorikian/commonmarker , which is a wrapper around comrak.

> I am not advising that. I am advising an allow list of http and https. I think that that is enough. For more complex needs, users can use their own sanitizing algorithm. Especially, when plugins are there.

I think this is a weaker security posture than allowing all safe protocols and forbidding unsafe ones by default. This encourages people who just need safe data URIs to switch on `allow_dangerous_protocol`, which will also enable `javascript:` URLs in links, which is a real problem.

---

Anyway, let's try to be productive here. I understand that you are not willing to allow `data:` URLs in images. If I cannot convince you that it's actually perfectly safe, then how can we proceed ? Would you be open to a pull request that makes the list of allowed protocols configurable, by storing `SAFE_PROTOCOL_SRC` in the configuration object instead of a hardcoded constant ?

wooorm:
> It does explicitly state that User agents must not run executable code (e.g. scripts) embedded in the image resource

Very nice.
Do you know when they added this?
As I have shown with Opera and DOMPurify, there have been used agents who did not follow this.
I can understand you only caring about modern browsers, but that is not necessarily what is best for everyone using this library + micromark by default.

> I think github uses https://github.com/gjtorikian/commonmarker , which is a wrapper around comrak.
Source: https://github.com/github/markup?tab=readme-ov-file#github-markup

😅 When GH adds things, they do it in `cmark-gfm`. Then it is alive here on GH. Such as footnotes a couple years ago and more recently https://github.com/github/cmark-gfm/commit/d0866dd92ed2086745ca30913a8147647d3a6fe0. You can observe the bugs between what is going on here in comments/files and what is in the code there.

> I think this is a weaker security posture than allowing all safe protocols and forbidding unsafe ones by default.

The latter is I believe what this project does. What is safe and unsafe, at what time, seems to be the debate.

***

There are indeed several discussions going on here.
One is about image formats potentially being insecure, scripts hidden in them.
One is about the `data:` protocol being safe.
One is about the value of `src` on `img` always being safe.
Most of your argumentation is about the third: you say everything is safe in images.
The solution you propose, is for the second, either configuring `SAFE_PROTOCOL_SRC` or allowing `data:` for everyone.

I think solutions need to match real problems. Be good to use and be hard to abuse.

Would other users allow particular protocols on images? Does it make sense for people to allow `tel:`, `ftp:`, `mailto:`? Which protocols would make sense? Is it easy for users to abuse such features and hurt themselves?

If users want that, would they not also want to change the protocols of `href` on `a`?

If users want all this, is that not what plugins are for? Should we not advance plugins?

Or, can we figure out that indeed in modern user agents `src` on images are all safe now, when that changed.
Add a feature for that: `allow_image_src` or so.
With docs referencing the HTML spec and links + dates to PRs (whatwg/html, Chromium) for when that changed

ChristianMurphy:
> One is about image formats potentially being insecure, scripts hidden in them.
> One is about the data: protocol being safe.
> One is about the value of src on img always being safe.

Another consideration is what standard implementations do.
We don't necessarily want to diverge from CM or GFM, this could also be a discussion that could be taken upstream.

>  Add a feature for that: allow_image_src or so.

I'd be more open to that approach

lovasoa:
I opened a separate pull request adding a crate feature: https://github.com/wooorm/markdown-rs/pull/165

wooorm:
This feature was released in [`1.0.0-alpha.23`](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.23)!

PR #165, opened 2025-02-26
new crate feature: allow_all_protocols_in_img

lovasoa:
test "unsafe" protocols on image sources here: https://codepen.io/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-/pen/wBvzLmv

This is a follow-up on https://github.com/wooorm/markdown-rs/pull/164#issuecomment-2682317922

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -290,6 +290,10 @@ protocols in links/images (such as `javascript:` or `data:`).
 dangerous protocols are used, as it encodes or drops them.
 Turning on the `allow_dangerous_html` or `allow_dangerous_protocol` options for
 user-provided markdown opens you up to XSS attacks.
+ - `allow_dangerous_html` allows HTML tags to be rendered, including tags that may trigger the execution of scripts
+ - `allow_dangerous_protocol` allows the use of protocols like `javascript:`, in links/images
+   - when the crate feature `allow_all_protocols_in_img` is enabled, `allow_dangerous_protocol` will only apply to links.
+     - The [HTML specification](https://html.spec.whatwg.org/multipage/images.html#images-processing-model) does not allow the execution of scripts in images, whatever the protocol they use. All modern browsers respect this.
 
 An aspect related to XSS for security is syntax errors: markdown itself has no
 syntax errors.
--- a/readme.md
+++ b/readme.md
@@ -297,10 +297,6 @@ prevent this, making the use of external images safe.
 
 Turning on the `allow_dangerous_html` or `allow_dangerous_protocol` options for
 user-provided markdown opens you up to XSS attacks.
- - `allow_dangerous_html` allows HTML tags to be rendered, including tags that may trigger the execution of scripts
- - `allow_dangerous_protocol` allows the use of protocols like `javascript:`, in links/images
-   - when the option `allow_any_img_src` is enabled, `allow_dangerous_protocol` will only apply to links.
-     - The [HTML specification](https://html.spec.whatwg.org/multipage/images.html#images-processing-model) does not allow the execution of scripts in images, whatever the protocol they use. All modern browsers respect this.
 
 An aspect related to XSS for security is syntax errors: markdown itself has no
 syntax errors.
--- a/readme.md
+++ b/readme.md
@@ -289,16 +289,16 @@ protocols in links/images (such as `javascript:`).
 `markdown-rs` makes any markdown safe by default, even if HTML is embedded or
 dangerous protocols are used, as it encodes or drops them.
 
-You should be able to set `allow_any_src` safely.
+Turning on the `allow_dangerous_html` or `allow_dangerous_protocol` options for
+user-provided markdown opens you up to XSS attacks.
+
+Additionnally, you should be able to set `allow_any_img_src` safely.
 The default is to allow only `http:`, `https:`, and relative images,
-which is what GitHub does.
-But it should be safe to allow any value on `src`.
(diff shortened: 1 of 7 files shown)


wooorm:
I appreciate where this is going! Thank you for your work!

Most things here are options here. What’s the reason you wrapped this in a feature? 🤔 I have used features here because this is a `no_std + alloc` crate by default.

Alternatively, for an option name, how about an option name: `allow_any_src`?

For the docs, then, for § Security, I am thinking of something like:

```markdown
You should be able to set `allow_any_src` safely.
The default is to allow only `http:`, `https:`, and relative images,
which is what GitHub does.
But it should be safe to allow any value on `src`.
The [HTML specification][whatwg-html-image] prohibits dangerous scripts in
images and all modern browsers respect this and are thus safe.
Opera 12 (from 2012) is a notable browser that did not respect this.

…

[whatwg-html-image]: https://html.spec.whatwg.org/multipage/images.html#images-processing-model
```

I tried to show why things are the way they are, and why, as you show, it should be safe to deviate from those defaults.

wooorm:
Sorry, I did not mean to imply a crate feature!

`allow_any_img_src` sounds great :+1: to me!

lovasoa:
I changed it to a configuration option instead of a crate feature.

I also updated the readme. I put the section about Opera above, since it's not related to the new `allow_any_img_src` option (opera executed scripts from images served over http as well).

You may also want to add a separate subsection about external images, warning about user tracking and CSRF attacks. Something  like

> Allowing external images still allows an attacker to perform an HTTP GET request
> using the user's stored cookies, IP address, user agent, referer, etc.
> This allows them to track the user, or perform [CSRF attacks](https://en.wikipedia.org/wiki/Cross-site_request_forgery).
> Depending on your threat model, you may want to store images on your own server.

readme.md
```diff
@@ -285,11 +285,22 @@ The following bash scripts are useful when working on this project:
...
+images vulnerable to XSS attacks.
+However, all modern browsers respect the HTML specification and 
+prevent this, making the use of external images safe.
+
 Turning on the `allow_dangerous_html` or `allow_dangerous_protocol` options for
 user-provided markdown opens you up to XSS attacks.
+ - `allow_dangerous_html` allows HTML tags to be rendered, including tags that may trigger the execution of scripts
+ - `allow_dangerous_protocol` allows the use of protocols like `javascript:`, in links/images
+   - when the option `allow_any_img_src` is enabled, `allow_dangerous_protocol` will only apply to links.
+     - The [HTML specification](https://html.spec.whatwg.org/multipage/images.html#images-processing-model) does not allow the execution of scripts in images, whatever the protocol they use. All modern browsers respect this.
```
wooorm, line 303: I am not sure we need to *explain* options here? There is very in-depth docs in other places?

readme.md
```diff
@@ -285,11 +285,22 @@ The following bash scripts are useful when working on this project:
...
-protocols in links/images (such as `javascript:` or `data:`).
+protocols in links/images (such as `javascript:`).
 `markdown-rs` makes any markdown safe by default, even if HTML is embedded or
 dangerous protocols are used, as it encodes or drops them.
+
+Some very old browsers (such as Opera 12 from 2012) did not respect the HTML
+specification and executed scripts in images, rendering sites with user-controlled
+images vulnerable to XSS attacks.
+However, all modern browsers respect the HTML specification and 
+prevent this, making the use of external images safe.
```
wooorm, line 296: I don’t think the phrasing is accurate.

I tracked down when this was added to whatwg/html: it’s at https://github.com/whatwg/html/commit/f47c41c53.

That’s a time (2008) when hixie was basically on his own, working on a hard fork, committing directly to main.
Here’s how that spec looked at the end of that month: https://web.archive.org/web/20180904063826/http://www.whattf.org/specs/web-apps/current-work/#scripting.
Note how it explicitly states in that link at 5.4 Scripting that scripting can happen in `src`. And at the top of the document, there’s a warning for user agents from following whattf/html.

To add, it is only in 2019 that whatwg/html became the sole HTML spec.

I think it can be expected that there are some user agents out there that do not follow this part of whatwg/html.

That is to say, it is clear that not all browser prevented scripting years ago. And, not all *existing* browsers follow whatwg/html.

So, I think your language is too strong when you say that “Opera did not respect the HTML specification”, because *the* HTML specification was by W3C;
I also very much worry about sweeping language such as “all modern browsers” and “making […] images safe”.
I prefer some of the wording from [my post yesterday](https://github.com/wooorm/markdown-rs/pull/165#issuecomment-2685357659), which is more about what the spec currently says, and using words like “should”.

wooorm (commented):
Thank you!

I think the Security section is not yet ready.

The rest of my suggestions are just some rephrasing and line wrapping.

The code looks :+1:

readme.md
lovasoa: ok! I removed it.

readme.md
lovasoa: I replaced the text with the exact wording from your earler comment.

wooorm (approved):

lovasoa:
Great, thank you ! Can you make a new release so that I can use it in my crate ?

wooorm:
Thank you! Released in [`1.0.0-alpha.23`](https://github.com/wooorm/markdown-rs/releases/tag/1.0.0-alpha.23)!

PR #170, opened 2025-03-09
Fix text not serializing correctly when having utf8 boundaries

### LINKED ISSUE #169: mdast_util_to_markdown panics while trying to format UTF text with trailing spaces
Hi there,

There is a bug somewhere in state.rs → container_phrasing(), which makes it panic with the message like "byte index 212 is not a char boundary" at line 245. I reproduced it in simple test: https://github.com/johnlepikhin/markdown-rs/commit/e4c451ad85a1a3348ea094bfa6dd10d0a4e12178. Please note that source markdown at lines 7-9 contains sequence:

```
<UTF-8 character> <SPACE> <SPACE> <NEWLINE>
```

bnchi:
Fixes #169

### DIFF
--- a/mdast_util_to_markdown/src/state.rs
+++ b/mdast_util_to_markdown/src/state.rs
@@ -242,7 +242,9 @@ impl<'a> State<'a> {
                     results.push(' ');
                     new_info.before = " ";
                 } else {
-                    new_info.before = &results[results.len() - 1..];
+                    if let Some(last_char) = results.chars().last() {
+                        new_info.before = &results[results.len() - last_char.len_utf8()..];
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


johnlepikhin:
Hi there,

Any chance to merge it?

wooorm:
I missed/forgot about this. Will get to it soon!

johnlepikhin:
I just found another case which is not resolved with suggested fix:

```rust
    let doc = "я_𝄞";
    assert_eq!(to(&from(doc, &Default::default()).unwrap()).unwrap(), doc);
```

```
---- roundtrip stdout ----
thread 'roundtrip' panicked at mdast_util_to_markdown/src/state.rs:540:51:
byte index 7 is not a char boundary; it is inside '𝄞' (bytes 4..8) of `
я_𝄞
```

wooorm:
Found them, thanks! :)

PR #181, opened 2025-06-16
fix: correct character indexing for non-ASCII strings in markdown escaping

CirnoV:
Fixes indexing issues when processing non-ASCII strings during markdown conversion.

- Replace `value.chars().nth(*position)` with `value[*position..].chars().next()` for proper byte-based
  indexing
- Prevents potential panics when handling multi-byte UTF-8 characters
- Update test cases


wooorm:
Thanks for your patience and the PR!

> Prevents potential panics when handling multi-byte UTF-8 characters

Did you get such a panic?
And, is it possible to add that case to the tests?

mdast_util_to_markdown/tests/paragraph.rs
```diff
@@ -114,7 +114,20 @@ fn paragraph() {
...
+            children: vec![Node::Text(Text {
+                value: String::from("테스트 [테스트] → [테스트]"),
+                position: None
+            })],
+            position: None
+        }))
+        .unwrap(),
+        "테스트 \\[테스트] → \\[테스트]\n",
+        "should support escaping of a shortcut reference in non-ascii text"
+    );
```
CirnoV, line 132: @wooorm Running this test on current main branch, it panics

image

PR #184, opened 2025-08-06
Add GFM Table Serialization Support to mdast_util_to_markdown

### LINKED ISSUE #171: Tables are not supported in mdast_util_to_markdown
Seems like the JS version has tables supported as extension: https://github.com/syntax-tree/mdast-util-gfm-table

Right now when you generate an AST and serialize that this error is reached: https://github.com/search?q=repo%3Awooorm%2Fmarkdown-rs%20unexpected-node&type=code

jp-knj:
Closes: https://github.com/wooorm/markdown-rs/issues/171
## Overview

This PR implements table serialization support for the mdast_util_to_markdown crate, bringing it to feature parity with the JavaScript https://github.com/syntax-tree/mdast-util-gfm-table package. Previously, attempting to serialize ASTs containing GFM table nodes would result in an "unexpected-node" error.

## Testing

Added comprehensive test coverage in `tests/table.rs`

- Basic table structures
- Column alignment variations
- Unicode content handling
- Pipe escaping scenarios
- Empty cells and edge cases
- Complex inline elements
- Error conditions

Run tests with:
```bash
# Build with the feature
cargo build --features unicode-width

# Run tests with the feature
cargo test --features unicode-width
```

Uses conditional compilation to:
- Default: Character count (`.chars().count()`), no dependency needed
- With feature: Unicode display width, proper CJK/emoji alignment

### DIFF
--- a/mdast_util_to_markdown/Cargo.toml
+++ b/mdast_util_to_markdown/Cargo.toml
@@ -1,7 +1,11 @@
 [dependencies]
 markdown = { path = "../", version = "1.0.0" }
 regex = { version = "1" }
-unicode-width = { version = "0.1" }
+unicode-width = { version = "0.1", optional = true }
+
+[features]
+default = []
+unicode-width = ["dep:unicode-width"]
 
 [dev-dependencies]
 pretty_assertions = { workspace = true }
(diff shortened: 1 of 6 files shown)


wooorm:
unicodewith seems like a step way from the upstream. And, a dependency.
Why not make it optional, configurable?

jp-knj:
You're right, I'll do that

PR #194, opened 2025-11-12
Fix nested brace counting in MDX expressions

rlch:
Remove parser check that prevented brace counting when external parser is configured. The original logic skipped counting opening braces in gnostic mode, causing expressions like `{{ x: 1 }}` to exit prematurely at the first closing brace.

Tokenizer must count all braces to find expression boundaries before the external parser validates syntax.

Fixes:
- Nested object literals: `{{ x: 1 }}`
- Nested function calls: `func({ x: g({ y: 2 }) })`
- JSX attribute expressions with nested objects

Tests added for nested object literal scenarios.


wooorm:
The idea is that if you have `mdx_expression_parse`, then you want something different than counting, because it doesn’t work, and instead accept/crash from `mdx_expression_parse` based on what language you expect.
Otherwise, why is `mdx_expression_parse` there?

So, why is `mdx_expression_parse` there? Why don’t you use the existing methods, from there, to signal back whether things are as expected or not?

wooorm:
It also looks like you want JS in there, which is exactly what is supported here? Why not use that? https://github.com/wooorm/markdown-rs/blob/1506572f9b406431402928f3a8b3df0b4ae2d8f5/tests/test_utils/swc.rs#L149-L156

PR #197, opened 2026-02-07
Fix scaling problems

### LINKED ISSUE #113: Performance: larger MDX files are unmanagably slow to parse
Bad news @wooorm - micromark/micromark#169, or something like it, is an issue here as well - I haven't yet grokked edit buffers, but either they don't solve the quadratic-complexity parsing problems or there's a separate performance bug in markdown-rs.

The constant factors are better, but the asymptotic complexity means that we're back to 60-second parse times on files that are just an order of magnitude bigger.

For comparison, the "JS" lines on these graphs show micromark's performance using subtokenize 2.0.1, which picks up the fix in micromark/micromark#171.

parse time (ms) vs  size(3)

log-log parse time (ms) vs  size(3)

### Data collection sources

Run with `node main.mjs` and `cargo run --release`

package.json
```
{
  "dependencies": {
    "mdast-util-from-markdown": "^2.0.0",
    "micromark-extension-mdxjs": "^3.0.0"
  }
}
```

main.mjs
```
import { fromMarkdown } from "mdast-util-from-markdown";
import { mdxjs } from "micromark-extension-mdxjs";

/* Generate test data */
const NUM_LINES_IN_CODE_SEG = 10000;
function generateTestData(size) {
  let file = [];
  for (let i = 0; i < size; i++) {
    file.push("", "<DummyComponent code={`");
    for (let j = 0; j < NUM_LINES_IN_CODE_SEG; j++) {
      file.push(crypto.randomUUID());
    }
    file.push("`} />", "");
  }
  return file.join("\n");
}

/* Return the number of milliseconds need to parse a test case */
function testPerformance(size) {
  const file = generateTestData(size);
  const start = performance.now();
  fromMarkdown(file, { extensions: [mdxjs()] });
  const end = performance.now();
  return end - start;
}

for (let reps = 0; reps < 3; reps++) {
  for (let size = 4; size < 40; size = Math.floor(size * 1.35)) {
    console.log(`${size}, ${testPerformance(size)}`);
  }
}
```

Cargo.toml
```
[package]
name = "tmp"
version = "0.1.0"
edition = "2021"

[dependencies]
markdown = "1.0.0-alpha.16"

[dependencies.uuid]
version = "1.8.0"
features = [
    "v4",                # Lets you generate random UUIDs
    "fast-rng",          # Use a faster (but still sufficiently random) RNG
]
```

src/main.rs
```
const NUM_LINES_IN_CODE_SEG: i32 = 10000;

fn main() -> Result<(), String> {
    for _ in 0..3 {
        let mut reps = 3;
        loop {
            reps = ((reps as f64) * 1.35) as i32;
            if reps > 40 {
                break;
            }
            let mut result = String::new();
            for _ in 1..reps {
                result.push_str("\n<DummyComponent code={`\n");
                for _ in 0..NUM_LINES_IN_CODE_SEG {
                    result.push_str(&uuid::Uuid::new_v4().to_string());
                    result.push('\n');
                }
                result.push_str("`}/>\n\n");
            }
            let start = std::time::Instant::now();
            let mdast = markdown::to_mdast(&result, &markdown::ParseOptions::mdx())?;
            let end = std::time::Instant::now();
            let duration = (end - start).as_millis();
            println!("{},{},{:?}", reps, duration, mdast.position());
        }
    }
    Ok(())
}
```

frankharkins:
I believe this fixes https://github.com/wooorm/markdown-rs/issues/113 by switching `EditMap` from a `Vec` to a `BTreeMap`.

I created a version of @robsimmons benchmark and added it to `benches`. Here's the result of running the `large_jsx_expressions` benchmark against the main branch (`Vec`) and my `BTreeMap` implementation (this PR), varying the `num_jsx_lines_per_component` variable.

plot


benches/bench.rs
```diff
@@ -1,24 +1,41 @@
...
-// , one_and_a_half_mb
+fn large_jsx_expressions(c: &mut Criterion) {
+    let num_components = 3;
+    let num_jsx_lines_per_component = 5000;
+    fn dummy_component(code: String) -> String {
+        return format!("<DummyComponent code={{`{code}`}} />");
+    }
+    fn uuids(len: usize) -> String {
+        (0..len)
+            .map(|_| "770f93e8-b4ee-4ce8-ab0f-4ece7d8c1090")
```
frankharkins, line 28: Hardcoding the same UUID seemed to reproduce the scaling behaviour, I left it like this to avoid adding another dependency.

benches/bench.rs
```diff
@@ -1,24 +1,41 @@
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use itertools::Itertools;
 use std::fs;
 
+fn tiny_markdown_string(c: &mut Criterion) {
+    let doc = "A *single* [markdown](/path) string!".to_owned();
```
frankharkins, line 6: I noticed a very small slowdown in this tiny test case (~8.9us -> ~9.1us; ~3%). This seems to even out very quickly as size grows. I tried creating an adaptive data structure that switched from `Vec` to `BTreeMap` at a certain size, but the complexity was quite high and the performance improvement was basically negligible.

benches/bench.rs
```diff
@@ -1,24 +1,41 @@
...
+    c.bench_with_input(BenchmarkId::new("tiny", "markdown_string"), &doc, |b, s| {
+        b.iter(|| markdown::to_html(s));
+    });
+}
+
 fn readme(c: &mut Criterion) {
     let doc = fs::read_to_string("readme.md").unwrap();
```
frankharkins, line 13: Criterion reported no difference for this benchmark.

src/util/edit_map.rs
```diff
@@ -76,36 +78,28 @@ impl EditMap {
     }
     /// Done, change the events.
     pub fn consume(&mut self, events: &mut Vec<Event>) {
-        self.map
-            .sort_unstable_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
```
frankharkins, line 80: `BTreeMap.iter()` is sorted already

src/util/edit_map.rs
```diff
@@ -115,34 +109,28 @@ impl EditMap {
...
-
     if remove == 0 && add.is_empty() {
         return;
     }
 
-    while index < edit_map.map.len() {
-        if edit_map.map[index].0 == at {
-            edit_map.map[index].1 += remove;
-
+    match edit_map.map.get_mut(&at) {
```
frankharkins, line 122: I think was the source of the quadratic behaviour; beforehand we were iterating through the `Vec` to find elements, which is $O(n)$. With `BTreeMap`, we can get an element in $O(log(n))$ time.

Murderlon:
I'm not an knowledgable enough in this codebase to have opinions but just to wanted to say this looks like a great optimization 🙏 

CI is still failing btw

frankharkins:
Other than the Clippy lint, I believe CI is failing because our version of `swc_common` uses `serde::__private`, which shouldn't have been relied on and has been renamed in more recent patch versions.

I tried upgrading `swc_common` but it required quite a few changes to the codebase, including changing user-facing error messages. Instead, I've manually edited the lockfile to use the version of `serde` that was available when this version of `swc_common` was released.

Phaqui:
What's the holdup on this? Is there anything I can do to help? For my use case, the runtime of a tool I have that checks links in markdown files, went from around 2 minutes, to 2 seconds, with this pr, with exactly the same behavior.

frankharkins:
Glad to hear it worked well! I imagine the maintainers are busy and this isn't the highest priority thing on their plate. 

To maintainers: If you have a working lockfile, feel free to push it over mine to make review easier.

Murderlon:
friendly ping @wooorm

frankharkins:
@wooorm gentle bump on this. Please let me know if there's anything I can do to make it easier to review.

PR #204, opened 2026-04-12
[DOCKER SEEDING] add helix files

Fatasia222:


ChristianMurphy:
No thanks, we don't want to be a part of your coding benchmark

PR #216, opened 2026-07-24
Improve source location lookup performance

marcoshernanz:
## Summary

Replace the linear scan in `Location::to_point` with `slice::partition_point`.

`indices` is sorted and the existing loop finds the first line-end index greater than `offset`, which is exactly the partition point for `<= offset`. This preserves the result while reducing each lookup from O(number of lines) to O(log number of lines). `partition_point` was stabilized in Rust 1.52, before this crate's Rust 1.56 minimum.

## Performance

I found this while profiling Next.js's native MDX compiler, which uses `mdxjs` and this crate. Environment: 16 vCPU / 32 GiB x86-64, Node 24.14.1, Next.js `0950e8ae05db20c33302f02257114f5b857edabc`. The frozen corpus was Next.js's 60,900-byte `docs/01-app/03-api-reference/02-components/image.mdx`.

A 16-process A/A panel (500 compiles/process, concurrency 4) measured an arbitrary-label wall difference of -2.32%, exact permutation p=0.362. A 32-process fixed-binary A/B panel with the same setup measured:

- wall time: 2012.244 ms -> 1906.879 ms, **5.24% faster**, block-stratified permutation p=0.000566;
- user CPU: 7704.901 ms -> 7270.089 ms, **5.64% lower**, p=0.000066;
- all four eight-run blocks improved wall time: 2.79%, 4.54%, 4.94%, and 8.52%;
- system CPU and RSS were neutral.

Matched 199 Hz `perf` profiles moved `Location::to_point` from 5.77% to 0.27% self CPU. Total sampled CPU fell from 31.452 s to 29.970 s.

## Correctness

- `cargo test --release`: full unit, integration, and doc-test suite passes (the published 1.0.0 lockfile was used because a fresh unlocked dev-dependency resolution currently fails on unchanged `main` in `swc_common`/`serde`).
- Compiled all 441 Next.js documentation MDX files with baseline and candidate under both CommonMark and GFM options: 882/882 outputs were byte-identical.
- Next.js focused MDX-rs integration suite: 4/4 passes under Turbopack and 4/4 under webpack.


marcoshernanz:
@wooorm @ChristianMurphy , could one of you review this narrow optimization or identify the right reviewer? It changes one lookup from O(lines) to O(log lines), preserves all 882 differential outputs, passes the Next.js MDX integration suite, and improves the representative Next.js compile by ~5%. The failing checks originate in untouched Clippy/dependency code; I can submit a separate CI repair first if preferred.

PR #217, opened 2026-07-24
Improve large edit map lookups

marcoshernanz:
## Summary

Use a `BTreeMap` index for `EditMap` position lookups after the map reaches 32 entries, while retaining the current linear scan for small maps.

The vector remains the source of truth and keeps the existing edit ordering and `consume` behavior. The index only maps source positions to stable vector offsets, is built at the threshold, updated on new entries, and discarded when `consume` resets the map.

## Motivation

A representative Next.js MDX document (`docs/01-app/03-api-reference/02-components/image.mdx`, 60,900 bytes) exposed one large edit map where repeated linear scans dominate `EditMap::add_impl`:

- 3,638 `add_impl` calls
- 1,184,540 vector entries scanned
- 3,606 lookup misses / 32 hits
- 324 consumed maps total
- 315/324 maps stayed at or below 32 entries
- the largest map had 1,427 entries and accounted for 1,017,451 scans (85.9%)
- the 9 maps above 32 entries accounted for 99.84% of all scans

This makes a thresholded index useful: the common small-map path is unchanged, while the long tail no longer performs quadratic search work.

## Benchmark

Environment:

- 16 vCPU / 32 GiB Vercel DevBox
- Node.js 24.14.1
- Rust nightly 2026-06-24
- identical source corpus, toolchain, process concurrency, and cache state for each pair
- exact output hash and length checked on every run
- balanced interleaved order, with a complementary second panel reversing assignment
- two-sided exact paired sign-flip p-values

### Standalone markdown-rs harness

Each sample compiled the 60.9 KB MDX input 200 times.

- A/A, 16 pairs: candidate-label delta `-0.496%`, `p=.566`
- A/B panel 1, 32 pairs: `+2.812%`, `p=.00285`; all four 8-pair blocks positive (`+4.947%`, `+4.251%`, `+0.003%`, `+1.968%`)
- A/B complementary panel, 32 pairs: `+1.946%`, `p=.000477`; all blocks positive (`+1.022%`, `+3.460%`, `+2.022%`, `+1.251%`)

### Next.js native binding

Each sample compiled the same input synchronously 200 times through the Next.js native binding.

- A/A, 16 pairs: wall delta `-1.915%`, `p=.097`; user CPU `-1.771%`, `p=.119`
- A/B panel 1, 32 pairs: wall `3010.185 -> 2823.479 ms` (`+6.202%`, `p<.000001`); user CPU `2910.910 -> 2736.330 ms` (`+5.997%`, `p<.000001`)
- A/B complementary panel, 32 pairs: wall `2888.470 -> 2743.180 ms` (`+5.030%`, `p<.000001`); user CPU `2814.734 -> 2672.929 ms` (`+5.038%`, `p<.000001`)
- all eight native 8-pair blocks were positive (`+3.129%` to `+8.506%`)
- output hashes and lengths were identical in every pair; peak RSS was neutral

Matched 1,000-compile `perf` profiles showed:

- total sampled CPU: `14.603 -> 13.729 s` (`5.98%` lower)
- `EditMap::add_impl`: `5.02% -> 0.59%` of samples
- indexed lookup itself was small (`BTreeMap::get`: `0.11%`)

## Correctness

- current upstream unit, integration, and documentation tests pass with the published `markdown` 1.0.0 lockfile
- 882/882 exact output comparisons pass across 441 real Next.js MDX files in both common-markdown and GFM modes
- focused Next.js MDX integration suite passes in both Turbopack (4/4) and webpack (4/4) modes
- `cargo fmt --check` and diff checks pass
- independent Codex Sol xhigh + Claude Opus xhigh pre-commit review found no actionable issues

This is independent from #216: that PR optimizes source-location lookup, while this one removes large-map edit lookup scans.


frankharkins:
I think this solves the same problem as #197, might be worth comparing the two. I added a benchmark in the PR.

If you do merge this, I think it'll probably fix #113.

PR #221, opened 2026-09-21
Remove unused extern crate std in gfm_tagfilter.rs

### LINKED ISSUE #220: crate uses `extern crate std` directly without a feature gate
This crate is marked `#![no_std]`, but it directly references an item from `std` in [util/gfm_tagfilter](https://github.com/wooorm/markdown-rs/blob/main/src/util/gfm_tagfilter.rs#L6) without any conditional compilation gate. This breaks `no_std` builds and prevents compilation on bare-metal or other `no_std`-only targets.

This seems to be an artifact from old debug sessions. Remove the line to fix the issue.

souragc:
Fix #220


ChristianMurphy (approved):

souragc:
@ChristianMurphy, do we need to fix the workflow issues before merging?

